Issue #85, opened 2015-01-27
Matching against empty body


algra:
I have Play application exposing an API, my controller looks like that

    class ApiController @Inject()(someClass: SomeClass) extends Controller {
      def addStuff = CorsAction.async(CustomBodyParsers.somePayloadFromJson) { request ⇒
        async {
          await(someClass.someMethod(someParams))
          Ok
        }
      }
    }

CorsAction is doing quite nothing except adding headers to allow cross-domain calls, removing it does not make any difference:

    object CorsAction extends ActionBuilder[Request] {
      val corsHeaders = Seq(
        "Access-Control-Allow-Origin" -> "*",
        "Access-Control-Allow-Methods" -> "POST, GET, PUT, DELETE, OPTIONS",
        "Access-Control-Allow-Headers" -> "Origin, X-Requested-With, Content-Type, Accept, Referer, User-Agent",
        "Access-Control-Max-Age" -> "1728000"
      )
      def invokeBlock[A](request: Request[A], block: (Request[A]) => Future[Result]) =  Async.async {
        await(block(request)).withHeaders(corsHeaders: _*)
      }
    }

This is how I'm matching response in pact file:

    {
      "description": "A POST request to add stuff",
      "request": {
        "method": "post",
        "path": "/add",
        "headers": {
          "Content-Type": "application/json"
        },
        "body": {
          "someName": "someValue"
        }
      },
      "response": {
        "status": 200
      }
    }

And pact fails with the following output:

    BodyMismatch(None,Some(),None,/) (Transformer.scala:20)

So, I guess None and empty body are not 100% correctly matched here: https://github.com/DiUS/pact-jvm/blob/master/pact-jvm-matchers/src/main/scala/au/com/dius/pact/model/Matching.scala#L101-L106

Can you, please, check it?

uglyog:
It is treating a missing body (i.e. no body attribute in the pact file) as different as an empty body. This should not be the case.

Changing the pact file to have an empty body may match. But I'll update the de-serialisation code to return a None for an empty body as well.

uglyog:
You should not get a failure now, as the empty response body should be mapped to None.

algra:
Yeah, I tried to put "body": "" in pact file, but it did not helped, mismatch was still there, don't have log output, but it was something like

Some("") vs Some()

Anyway, should be fixed now with your fix, thanks a lot for a quick one :) Do you have plans to release a new version to maven repo?

algra:
[...]

PR #87, opened 2015-01-29
Fixing the problem of evaluating mime type even if no body is specified....

algra:
... In this case "fallback" mime type "text/plain" is used, which makes test failing if provider returns some json or xml. #85.


uglyog:
Thank you for that.

algra:
Can you, please, release new version with fix of matching of empty body, so we can enable pact testing in our pipeline? Thanks!

uglyog:
I'll push a new version to maven central tonight (Australian standard eastern time, +11).

PR #91, opened 2015-02-22
Test without pact verification still gets run as a pact consumer

hbakkum:
I have an integration test class that includes tests both with and without pact verification. Without this change, I cannot get the tests with no pact verification to pass as I get a NPE on line 64 of PactRule. I see a guard is in place to execute tests without pact verification normally, but without the return, the rule continues to execute the test as a pact consumer.


uglyog:
Thanks for that

PR #94, opened 2015-03-02
README uses markdown to link to pact specification

nicknovitski:
I myself mix up round and square brackets in markdown links on a regular basis.


uglyog:
Yeah, damn brackets. They catch me out all the time! ;-)

PR #95, opened 2015-03-05
Make 'findHeaderByCaseInsensitiveKey' public so user can use it to find a header easily.

freewind:
Because in the https://github.com/realestate-com-au/pact-jvm-provider-spring-mvc project, I need to get the cookie value from the request, and don't want to implement a duplicate method like `findHeaderByCaseInsensitiveKey` again


uglyog:
merged

Issue #83, opened 2015-01-23
Defining (HTTP-Request) content type for url based pactFile in PactVerificationTask


hendrikstill:
If I want to use a external pactFile from an HTTP-Server to verify my Provider with the PactVerificationTask my gradle file looks like this 

```groovy
pact {
	serviceProviders {
		productPriceServiceProvider {
			protocol = 'http'
			host = 'localhost'
			port = 10101
			path = '/'
			hasPactWith('productServiceConsumer') {
				pactFile = url("http://localhost:9292/pacts/provider/Product_Price_Service/consumer/Product_Service/latest")
			}
		}
	}
}
```

This works fine, but if you try to access url's which expect a specific Header (e.g. Accept: 'application/json'), the Task will not fetch the correct content.
This is the case if you use the [Pact Broker](https://github.com/bethesque/pact_broker). 
The request of a pact e.g. "http://localhost:9292/pacts/provider/Product_Price_Service/consumer/Product_Service/latest" without the  (Accept: 'application/json') will return a HTML-Page.

The request parameters should be passed into this line [1] as parameter of  `newInputStream`  method (see [2]).

[1]https://github.com/DiUS/pact-jvm/blob/master/pact-jvm-provider-gradle/src/main/groovy/au/com/dius/pact/provider/gradle/PactVerificationTask.groovy#L34
[2]http://groovy.codehaus.org/groovy-jdk/java/net/URL.html#newInputStream%28java.util.Map%29

uglyog:
I've added an accept header to the URL fetch. Hope this fixes your issue.

bethesque:
This does not sound right. The default content type returned by the pact broker for a pact resource is application/json. It will only return a HTML page if the Accept header includes HTML. I'd suggest that the client that is making the request might be adding an HTML Accept header by default if one is not specified.

hendrikstill:
@bethesque you are right, with curl I get the correct response. 
So as correction: the Pact Broker returns a JSON, if no Accept Header is given.
But I will validate if 4ddb281 fixes my problem

gthicks:
This change doesn't appear to have solved the problem. I looks like the accept header is not being set correctly. Using a packet sniffer, I found that requests are being made with "Accept: text/html, image/gif, image/jpeg"

I think that the proper syntax for setting the request headers is:
```
pact = Pact$.MODULE$.from(new StreamInput(consumer.pactFile.newInputStream(requestProperties: ['Accept': 'application/json'])))
```

[...]

PR #99, opened 2015-03-16
properly set accept type to ‘application/json’

gthicks:
address issue #83


gthicks:
I looks like the build failed due to memory issues. Is there a way to  re-run the build in Travis?

uglyog:
I'll kick the build off again

uglyog:
That looks correct occording to the docs (don't knowhow I missed that!) Also, as two of the builds have passed, I'm going to merge it.

PR #103, opened 2015-03-24
Allow PactDslJsonArray to be at the root of a response

jamessawle:
Allow PactDslJsonArray to be at the root of a response


jamessawle:
Hi, this is my first pull request to this project. I have not changed anything that would have affected the building of the scala modules. Any advice on how to rectify this?

uglyog:
Awesome, thanks for that!

Issue #104, opened 2015-03-25
'Template Objects' in PactDslJsonBoday and PactDslJsonArray


jamessawle:
Currently, each Pact must explicitly define each part of its expected response. Whilst this gives clarity on exactly what is expected, this can make them difficult to read when there are repeated Objects in a response, or even across responses. Especially when one of the 'template objects' has 30 fields and nested arrays as well'

As such it would be nice to make the protected 'putObject' methods public on both the PactDslJsonBody and the PactDslJsonArray. On the array, it would also be beneficial to be able to have a second wrapper method that allowed for the object to be added n times.

Any ideas would be much appreciated, and I will prepare a pull request.

uglyog:
Closing, as pull request was merged

PR #105, opened 2015-03-25
https://github.com/DiUS/pact-jvm/issues/104 - Added method to allow for ...

jamessawle:
https://github.com/DiUS/pact-jvm/issues/104 

DslParts can now be used as templates within an array object for easier composition of contracts.


uglyog:
Merged, thanks for that.

Issue #106, opened 2015-04-01
Template method within PactDslJsonArray not generating matchers within an array


jamessawle:
The new 'template' methods on PactDslJsonArray are not placing the matchers within an array (sorry for that), causing Pacts not to match.

Will have a look tonight to see if I can come up with a quick fix for this; but may need some help :s

uglyog:
oh, yeah. Oops. Sorry, forgot about the matchers.

jamessawle:
Thanks, the fix solves the issue. Can not wait to simplify our CDC generation code with the next release :D

PR #107, opened 2015-04-06
Fix matcher name issue with templates in arrays

jamessawle:
https://github.com/DiUS/pact-jvm/issues/106 - Generate new matcher names as objects are placed into the array. Any common prefix should be able to be removed and replaced by the new root.

### DIFF
--- a/pact-jvm-consumer-junit/src/main/java/au/com/dius/pact/consumer/PactDslJsonArray.java
+++ b/pact-jvm-consumer-junit/src/main/java/au/com/dius/pact/consumer/PactDslJsonArray.java
@@ -2,6 +2,7 @@ package au.com.dius.pact.consumer;
 
 import nl.flotsam.xeger.Xeger;
 import org.apache.commons.lang3.RandomStringUtils;
+import org.apache.commons.lang3.StringUtils;
 import org.apache.commons.lang3.time.DateFormatUtils;
 import org.apache.commons.lang3.time.FastDateFormat;
 import org.json.JSONArray;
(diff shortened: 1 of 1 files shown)


jamessawle:
A simpler alternative could be to just strip '$.body' off the front of all of the matcher names, would allow for nested structure.

bethesque:
Is this an implementation detail, or changing the JSON serialisation?

uglyog:
We need the `$.body` because matchers could also be applied to other parts, like headers, etc.

jamessawle:
Beth - this is an implementation detail, trying to make the generation of complex json return objects simpler. The bug (that I introduced and am trying to fix) is that the matchers within the pact were not storing the index of the element in the array and therefore could never pass.

Uglyog - does it make sense to pass through header matchers in an object if just an element within an array? My concern was, that if you were to pass in an object with just an array with elements into my implementation, the out element and array would be lost and the elements of the inner array would be placed into the outer array as individual elements. If you were to just remove the $.body from the template object, the full structure of the template object would be preserved and reduce the overhead at the same time. Would probably have to change the signature to only take the JSON implementations though.

uglyog:
We could leave the `$.body` root out of the PactDslJsonArray and PactDslJsonBody, but we would need to add it as the root in ConsumerPactBuilder body methods. I actually like this idea, because the ConsumerPactBuilder knows that it is being added to body, so can set the root appropriately.

jamessawle:
That would work nicely until you have an array at the root of the

jamessawle:
Sorry accidentally hit comment

That would be a nice solution, but would require the 'root' of every object to end with the leading '.' of its parent, and therefore the root objects to just have a '.'

This would allow for the array index to be added in front of the '.' for each object in an array.

uglyog:
I guess you could check if it is an array, and set the root to `$.body` otherwise set it to `$.body.`.

The other option is to have arrays have an empty root, and objects have a root of `.`, then the builder can just add `$.body`. I'll need to think about this if this will work with nested arrays and objects.

jamessawle:
I will leave the patch for now then. Sorry :p

uglyog:
No prob. I think I've implemented the correct solution (commented against the original issue). If you can check it out to see if it resolves you issue, that will be appreciated.

I'll close this pull request in the mean time.

Issue #108, opened 2015-04-08
Version 2.1.13 of pact-jvm-provider-maven_2.11 does not work if it is configured to use pactUrl


GorshSerginio:
Version 2.1.13 of pact-jvm-provider-maven_2.11 fails with the following exception if `serviceProvider` is configured to use `pactUrl`:

```
org.apache.maven.lifecycle.LifecycleExecutionException: Failed to execute goal au.com.dius:pact-jvm-provider-maven_2.11:2.1.13:verify (default) on project account: Execution default of goal au.com.dius:pact-jvm-provider-maven_2.11:2.1.13:verify failed: Cannot invoke method newInputStream() on null object
	at org.apache.maven.lifecycle.internal.MojoExecutor.execute(MojoExecutor.java:224)
	at org.apache.maven.lifecycle.internal.MojoExecutor.execute(MojoExecutor.java:153)
	at org.apache.maven.lifecycle.internal.MojoExecutor.execute(MojoExecutor.java:145)
	at org.apache.maven.lifecycle.internal.LifecycleModuleBuilder.buildProject(LifecycleModuleBuilder.java:116)
	at org.apache.maven.lifecycle.internal.LifecycleModuleBuilder.buildProject(LifecycleModuleBuilder.java:80)
	at org.apache.maven.lifecycle.internal.builder.singlethreaded.SingleThreadedBuilder.build(SingleThreadedBuilder.java:51)
	at org.apache.maven.lifecycle.internal.LifecycleStarter.execute(LifecycleStarter.java:120)
	at org.apache.maven.DefaultMaven.doExecute(DefaultMaven.java:355)
	at org.apache.maven.DefaultMaven.execute(DefaultMaven.java:155)
	at org.apache.maven.cli.MavenCli.execute(MavenCli.java:584)
	at org.apache.maven.cli.MavenCli.doMain(MavenCli.java:216)
	at org.apache.maven.cli.MavenCli.main(MavenCli.java:160)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:57)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:606)
	at org.codehaus.plexus.classworlds.launcher.Launcher.launchEnhanced(Launcher.java:289)
[...]

PR #109, opened 2015-04-08
#108 - Fixed reference to pactUrl

GorshSerginio:


uglyog:
Good pick-up, thanks!

PR #125, opened 2015-05-31
Implement missing parts of XmlBodyMatcher

flosell:
The `XmlBodyMatcher` is currently disabled because it's incomplete. So here's a pull request to make it more complete and hopefully ready for the next release.  
I might not have covered all special cases but I hope this implementation is a start. 

Regarding the paths I'm not quite sure: At the moment, I'm sticking with the JSONPath syntax the `JsonBodyMatcher` is using since the Matchers seem to require it. It feels a bit strange for XML though and I'm not sure if it's expressive enough.


uglyog:
That's awesome. Thank you. XML matching has been outstanding for quite a while.

bethesque:
Can we get some pact-spec happening here? We haven't really discussed the logic for the XML matching, can we have some test cases? https://github.com/bethesque/pact-specification

Issue #145, opened 2015-07-27
Allow array matchers to work when root of the body is an array


batasha:
Currently, array matchers assume that there is an object at the root of the body rather than an array, eg. 
```
{
  "values": [
    {
      "clearedDate": "07/22/2015",
      "amount": 15.0
    },
    {
      "clearedDate": "07/22/2015",
      "amount": 15.0
    }
  ]
}
```
I would like to be able to use this feature when the root of the body is an array, eg.
```
[
    {
      "clearedDate": "07/22/2015",
      "amount": 15.0
    },
    {
      "clearedDate": "07/22/2015",
      "amount": 15.0
    }
]
```
Here's the link to the corresponding Google thread: https://groups.google.com/forum/#!topic/pact-support/T1VhZrm4OJg

batasha:
In my case I'm using the JUnit consumer, but it looks like someone posted a similar issue specifically for groovy: https://github.com/DiUS/pact-jvm/issues/141

uglyog:
Thanks for that.

#141 is related, but more to do with the Groovy DSL not supporting array bodies at all.

uglyog:
I've added a static method to PactDslJsonArray to help. You can use it like:

```java
PactDslJsonArray.arrayEachLike()
            .date("clearedDate", "mm/dd/yyyy")
            .stringType("status")
            .realType("amount")
            .closeObject();
```

This will create an array matcher you can use as the root of the body.

batasha:
Thanks so much for looking into this! I noticed the matchers were missing from the arrayEachLike method, so I just put in a PR to fix that.

PR #146, opened 2015-08-01
Add missing matchers to arrayEachLike() #145

batasha:
I saw that the matchers were missing from the new arrayEachLike() method, so I just added them in.


uglyog:
Good catch, thanks for that.

PR #162, opened 2015-09-15
Scan classpath is not set in pact-jvm-provider-maven v3 pact verifier

venksub:
When the v3 pact verifier runs, it needs to scan the runtime classpath for annotated methods. In the maven plugin, the plugin code does not set the projectClassPath which causes an error when the verifier runs.

Sorry, I am not very good with Git and somehow, two commits were pulled into this request. You only need the "Fix for no project classpath in maven v3 pact verifier" commit


uglyog:
Thank you for that!

PR #164, opened 2015-09-21
Java DSL and JUnit infrastructure implementing V3 specification changes.

venksub:
Mostly a copy over of the groovy v3 consumer side package. I made one change which is to set the message bytes into an instance variable of the test class. This makes it easy for test code to pull the message and test the contents (say as JSON).


uglyog:
Awesome, thank you!

PR #181, opened 2015-11-09
Add ability to play contract tests against provider service in JUnit fashionable way.

knok16:
Incorporate pact-provider-testing with JUnit: screenshot


uglyog:
Nice! This is really useful. I'll just do a quick code review and then I can merge it in.

PR #193, opened 2015-12-03
add @PactUrl to pact-jvm-provider-junit

stefan-lz:
Please review this and merge into master for me

Cheers,

### DIFF
--- a/pact-jvm-provider-junit/README.md
+++ b/pact-jvm-provider-junit/README.md
@@ -63,6 +63,10 @@ Pact runner will automatically collect pacts: for this purpose there are 2 out-o
 
 To use pacts from Pact Broker annotate test class with `@PactBroker(host="host.of.pact.broker.com", port = 80)`.
 
+### Pact Url
+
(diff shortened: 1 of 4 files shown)


uglyog:
I've merged it, but I'm missing seeing the link between the annotation and the loader. What handles calling the loader when the annotation is present?

idntfy:
That's [handed by infrastructure](https://github.com/DiUS/pact-jvm/blob/3_2_1/pact-jvm-provider-junit/src/main/java/au/com/dius/pact/provider/junit/PactRunner.java#L94) that has a pact source abstraction. PactUrl is just another pact source, like [PactBroker](https://github.com/DiUS/pact-jvm/blob/3_2_1/pact-jvm-provider-junit/src/main/java/au/com/dius/pact/provider/junit/loader/PactBroker.java) or [PactFolder](https://github.com/DiUS/pact-jvm/blob/3_2_1/pact-jvm-provider-junit/src/main/java/au/com/dius/pact/provider/junit/loader/PactFolder.java).
Basically anyone can add any pact source. For example, you can have even your company-specific `@MyCompanyPactSource` in order not to hardcode pact broker url everywhere.

uglyog:
Nice!

PR #203, opened 2015-12-23
Added one line description of ConsumerService to README example that …

davesmith00000:
…would have saved me 20 minutes...

It's a small thing but I think new comers expect that, once you add the library dependency, you should be able to copy and paste example code with an expectation of it compiling, or an explanation for why it won't. :-)


uglyog:
Thank you. I been getting a lot of questions about that and hopefully this will make it easier to understand.

PR #206, opened 2016-01-11
Update net.databinder.dispatch version to 0.11.3

treppo:
Some method signatures changed, so I had to adapt the calls. Is there anything else I need to take care of?


treppo:
It's weird that the Travis CI build fails, I don't see any errors and locally the build command used by Travis `./gradlew --stacktrace --no-daemon -i check_2.11` does not return successfully.

uglyog:
This is an issue with Travis and Gradle. The build actually passed, you can see the following at the end of the raw log:

```
:check_2.11 (Thread[main,5,main]) completed. Took 0.0 secs.

BUILD SUCCESSFUL

Total time: 9 mins 18.47 secs

FAILURE: Build failed with an exception.

* What went wrong:
Process 'Gradle Compiler Daemon 2' finished with non-zero exit value 137
```

From what I can work out, Travis is killing the Gradle process before it can shutdown properly.

treppo:
OK, good to know. Thanks!

PR #214, opened 2016-02-12
bugfix - nested JSON objects get parent nodes prefixed

tobivogel:
With my last pull request - fixing the matchers for nested objects - i introduced a bug that lead to the output of wrong JSON objects where nested objects got the ids of their preceding objects appended (as matchers do).
This will be fixed with this PR.

...please excuse the inconvenience. =o)


uglyog:
Thank you for the followup. Don't worry about any inconvenience, nobody has reported an issue yet an I will push out an update release.

PR #225, opened 2016-03-15
@PactVerifications

andreschaffer:
The pact-jvm-consumer-junit module currently does not support to have both an http and a message pact in the same test class using the PactProviderRule and MessagePactProviderRule (an exception is thrown).
Also, the @PactVerification annotation allows to specify multiple providers but only one fragment, thus not supporting the map of multiple providers with specific fragments each.

I implemented a new annotation, @PactVerifications, that aims to solve these two issues. For that to work, I also added the annotation handling in both the PactProviderRule and the MessagePactProviderRule, while keeping the old behavior for when the @PactVerification is used alone.


uglyog:
Thanks for that, looks extremely useful.

andreschaffer:
Glad to help! :)

PR #228, opened 2016-03-22
fixes toString problem in PactDslResponse when DslPart used

lordofthejars:
Fixes toString problem in PactDslResponse when DslPart used.

Basically if for example you do something like:

```
PactDslJsonRootValue.stringMatcher("[0-9]+.[0-9]", "123.2");
```

since `PactDslResponse` calls `toString` directly, if the method is not implemented then a weird string representation is shown instead of the value set as example. This is the case when `PactDslJsonRootValue`. Since `toString` is not overriden you cannot use the example value (123.2). And the same would happens with any `DslPart` object not implementing `toString` method. To mitigate this, the `getBody` is used to get the value.


uglyog:
Good catch!

lordofthejars:
You are welcome. A pleasure to help on this awesome project. I am going to show it briefly in Greach conf :) http://greachconf.com/ among other tools for doing CD for microservices.

mefellows:
@lordofthejars Great stuff! Will the conference be recorded? Let us know if there is anything we as a community can do better to engage and promote - feedback always welcome!

PR #251, opened 2016-04-25
bugfix - fix handling nested properties in PactDslJsonBody

awojcicki:
missing initial dot in matching key when property was escaped


uglyog:
Thank you for that fix!

PR #265, opened 2016-06-16
Add new line after a message 'Consumer 'X' has a pact with Provider 'Y''

olga-vasylchenko:
Hi,

I am using specs2 implementation on my project.
It works quite well, except there is one tiny detail that could improve the console output for the test results for me.
Here is the screenshot on how the output looks for me now (I have multiple assertions):
screen shot 2016-06-16 at 10 41 28

I would like '**_has valid price_**' in my example to move to a new line (as well as the other assertions).

Thanks!


uglyog:
Merged, thank you!

BTW, there is a Scala specific Pact library you might want to look at: https://github.com/ITV/scala-pact

PR #274, opened 2016-06-20
Allow HttpTarget to set insecure flag

### LINKED ISSUE #273: JUnit HttpTarget does not allow setting insecure flag
`HttpTarget` doesn't have a constructor option to pass an insecure flag to the `ProviderInfo` object created in `getProviderInfo`.  It should probably be added as an overloaded constructor parameter which is secure by default.

Fitzoh:
Overloads constructors to allow HttpTarget to disable cert checking/hostname verification.
Implemented by overloading constructors, still secure by default.
Also updated some of constructor javadocs for consistency.

Fixes #273


uglyog:
Thank you for the PR.

Fitzoh:
Quite welcome

PR #280, opened 2016-06-24
Better specs2 states support

alonpeer:
This PR:

- Adds support for switching between states for different provider requests.
- Fixes a bug where the value `providerState` in `PactSpec` was not applied.


alonpeer:
The build fails for unrelated tests.

uglyog:
I re-ran the build, and all passed. Thanks for the PR.

PR #286, opened 2016-07-09
Prevent recursion into the body elements of the pact file

whahoo:
Hi,
I had an issue where, while the pact file was being verified, it was uppercase-ing an element in the request body, the element was named _method_.
The `recursiveTransformJson` method, runs over the whole pact including request and response bodies, this means that if the element names being transformed here appear in the request or response they will be incorrectly transformed.

Please consider this fix for this issue.
Thanks,
Simon


uglyog:
Thanks for the PR! 

I might move the uppercasing of the method out of the recursive call as it does not need to be there.

uglyog:
I've actually removed the recursion all together as it did not make much sense.

whahoo:
Fair enough, I wasn't going to be so bold!
Thanks for looking into it :)

PR #289, opened 2016-07-21
Add support for character class shortcuts in Xeger

### LINKED ISSUE #288: Randomly generated matcher value does not match regex
I'm using `PactDslJsonBody` in Scala with the specs2 consumer. I'm trying to use a digit matcher like so:

```scala
val body = new PactDslJsonBody().stringMatcher("foo", """bar\d+""")
```

At runtime I get the following error:

```
...
[error]   CAUSED BY au.com.dius.pact.consumer.InvalidMatcherException: Example "bard" does not match regular expression "bar\d+"
[error]   au.com.dius.pact.consumer.dsl.PactDslJsonBody.stringMatcher(PactDslJsonBody.java:259)
[error]   au.com.dius.pact.consumer.dsl.PactDslJsonBody.stringMatcher(PactDslJsonBody.java:273)
...
```

I can currently bypass this by forcing a specific value:

```scala
val body = new PactDslJsonBody().stringMatcher("foo", """bar\d+""", "bar123")
```

However I prefer using the random values, so I can use this with arrays and get different values for each element.

alonpeer:
Fixes #288 

Code shamelessly copied from [Generex](https://github.com/mifmif/Generex/blob/master/src/main/java/com/mifmif/common/regex/Generex.java#L62-L71).

I've decided to just copy the relevant code from Generex, instead of introducing a new dependency in the Build. If you prefer otherwise, I have a branch for that which uses Generex instead of Xeger: https://github.com/alonpeer/pact-jvm/tree/alt-regex-gen

Also notice the last commit in this PR which deals with this issue I've found: https://github.com/mifmif/Generex/issues/26
This is reproducible with both Xeger and Generex. I couldn't find a way to resolve it so far. In any case, this has nothing to do with the character class shortcuts, which weren't supported correctly so far anyway.

### DIFF
--- a/pact-jvm-consumer-junit/src/test/java/au/com/dius/pact/consumer/Defect215Test.java
+++ b/pact-jvm-consumer-junit/src/test/java/au/com/dius/pact/consumer/Defect215Test.java
@@ -51,7 +51,7 @@ public class Defect215Test {
         .matchHeader(CONTENT_TYPE, APPLICATION_JSON, APPLICATION_JSON_CHARSET_UTF_8)
       .willRespondWith()
         .status(201)
-        .matchHeader("Location", "http(s)?://\\S+:\\d+//some-service/user/\\S{36}$")
+        .matchHeader("Location", "http(s)?://\\w+:\\d+//some-service/user/\\w{36}$")
       .given("An automation user with id: " + EXPECTED_USER_ID)
       .uponReceiving("existing user lookup")
         .path(SOME_SERVICE_USER + EXPECTED_USER_ID)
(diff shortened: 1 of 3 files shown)


alonpeer:
I have no idea why this fails in AppVeyor, it works for me locally.

uglyog:
I've kicked the AppVeyor again, but travis passes so I'm happy to merge.

uglyog:
Thanks for the PR.

I would prefer a build dependency. The only reason I inlined Xeger is because there was no published jar at the time.

alonpeer:
Thanks for merging it. As I mentioned, if you prefer a dependency, there's this PR: https://github.com/alonpeer/pact-jvm/tree/alt-regex-gen
From what I could tell, Generex and Xeger are quite similar.

PR #292, opened 2016-07-22
Add support for generating specs2 Pact with DslPart

alonpeer:
I hope I'm using it right 😄


uglyog:
Thanks for the PR! Looks good to me. I might modify it slightly, because you're change is converting the matchers to Scala collections, and then calling the previous method which converts it back again.

alonpeer:
Oh, good point. Thanks.

PR #294, opened 2016-07-23
Add scalatest ProviderSpec support

lachatak:
Hi,
I have added a scalatest ProviderSpec support to be able to test pacts in scalatest. 
pact-jvm-provider-specs2 project was used as an example but this scalatest version has a bit wider DSL and functionality support.
Tests and documentation clearly explains how it works.
I hope you will like it. Feel free to comment or suggest.
Krisztian


uglyog:
Thank you for the PR! That is awesome.

PR #302, opened 2016-08-05
Provides a way to set the body to 'missing' in PactFragmentBuilder

### LINKED ISSUE #248: Ignore body when validating
I've got a consumer that when requesting something by ID from a provider, I expect a 404 returned when it doesn't exist. I don't care about the body, purely the status code.

However, in version 2.4.8, I get an issue where the pact test fails, as it expects no body, but one is returned. This doesn't happen in previous versions of the library.

Is there a way around this?

alonpeer:
Fixes #248


uglyog:
Great, thanks for the PR.

PR #303, opened 2016-08-05
Add the $.body prefix to DslPart matchers

### LINKED ISSUE #300: Missing $.body when fetching matchers from DslPart
When running `au.com.dius.pact.consumer.specs2.ExamplePactWithMatchersSpec`, the outputted pact file has the following:

```json
            "response": {
                "status": 200,
                "body": {
                    "bar": "A",
                    "foo": "04"
                },
                "matchingRules": {
                    ".bar": {
                        "regex": "[aA]+"
                    },
                    ".foo": {
                        "regex": "\\d{1,9}"
                    }
                }
            }
```

The matching rules should have the keys `"$.body.bar"` and `"$.body.foo"`.

alonpeer:
Fixes #300 

I'm not entirely sure that this is the right solution. I had to alter some unit tests to make this pass, but perhaps I got it wrong, so please review this.

This change actually limits `DslPart` to only be used for creating matchers for `body`. Perhaps we need to generalize it so it can also be used for `header` and other HTTP parts (see [V2 specs](https://github.com/pact-foundation/pact-specification/tree/version-2#matcher-path-expressions)).

### DIFF
--- a/pact-jvm-consumer/src/test/groovy/au/com/dius/pact/consumer/PactDslJsonArrayMatcherSpec.groovy
+++ b/pact-jvm-consumer/src/test/groovy/au/com/dius/pact/consumer/PactDslJsonArrayMatcherSpec.groovy
@@ -50,10 +50,10 @@ class PactDslJsonArrayMatcherSpec extends Specification {
           [amount: 100, clearedDate: date.format('mm/dd/yyyy'), status: 'STATUS']
         ]
         subject.matchers == [
-          '': [min: 0, match: 'type'],
-          '[*].amount': [match: 'decimal'],
-          '[*].clearedDate': [date: 'mm/dd/yyyy'],
-          '[*].status': [match: 'type']
+          '$.body': [min: 0, match: 'type'],
+          '$.body[*].amount': [match: 'decimal'],
+          '$.body[*].clearedDate': [date: 'mm/dd/yyyy'],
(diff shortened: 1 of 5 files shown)


pact-jvm-consumer/src/test/groovy/au/com/dius/pact/consumer/PactDslJsonArrayMatcherSpec.groovy
```diff
@@ -50,10 +50,10 @@ class PactDslJsonArrayMatcherSpec extends Specification {
           [amount: 100, clearedDate: date.format('mm/dd/yyyy'), status: 'STATUS']
         ]
         subject.matchers == [
-          '': [min: 0, match: 'type'],
-          '[*].amount': [match: 'decimal'],
-          '[*].clearedDate': [date: 'mm/dd/yyyy'],
-          '[*].status': [match: 'type']
+          '$.body': [min: 0, match: 'type'],
```
alonpeer: Is this the right output to expect?

uglyog: :+1:

uglyog:
Looks all good to me, thanks for the PR. I think this is an example of something ending up after a series of refactors over time.

PR #309, opened 2016-08-11
Fix error message shown when @Pact is used incorrectly

carwinz:
When a method annotated with @Pact doesn't have the correct method signature, an error message is shown.

This corrects the error message


uglyog:
👍

PR #317, opened 2016-09-15
Created a VersionedPactUrl to be used to reference versioned pacts.

wojciechbulaty:
We are using this class internally for one of my clients. Let me know what you think. Thanks! :)


uglyog:
Thanks! That will be useful.

Issue #313, opened 2016-08-30
Wildcards in keys


felixhammerl:
There should be a way to use wildcards on JSON object's keys, as it is possible for values (e.g. strings). 

[Twitter convo w/ @uglyog on this topic](https://twitter.com/felixhammerl/status/770600678288949249)

### Background

This is a part of the JSON API we're writing a CDC for:

```
"articles": [
    {
        "variants": {
            "001": {
                "bundles": {
                    "001-A": {
                        "description": "someDescription",
                        "referencedArticles": [
                            {
                                "bundleId": "someId"
                            }
                        ]
                    }
                }
            }
        }
    }
]
```

There's one or more objects being used like an array (`001`, `002`, ... and `001-A`, `001-B`, ...). Not pretty, but not possible to change that.

I need something like this: `$.body.articles[*].variants.*.bundles.*.description`. Pact-JVM seems to allow wildcards (see [README](https://github.com/DiUS/pact-jvm/wiki/Matching#matcher-path-expressions) ), but I've not been able to do that in Java via the API. Also, I'm not trying to manually add it to the JSON as it is auto-generated. But the only API that does anything like this is for arrays (`.eachLike()`, etc.).

uglyog:
I have updated the DSLs to support wildcard keys, but it is not going to be pretty ;-)

To match you JSON requires something like:

```java
    DslPart body = new PactDslJsonBody()
        .eachLike("articles")
          .object("variants")
            .eachKeyLike("001")
              .object("bundles")
                .eachKeyLike("001-A")
                  .stringType("description", "Some Description")
                    .eachLike("referencedArticles")
                      .id("bundleId", 23456L)
                      .closeObject()
                    .closeArray()
                  .closeObject()
                .closeObject()
              .closeObject()
            .closeObject()
          .closeObject()
        .closeArray();
```

and this generates a pact file with the following body and matchers:

```json
        {
                "body": {
                    "articles": [
                        {
                            "variants": {
                                "001": {
                                    "bundles": {
                                        "001-A": {
                                            "description": "Some Description",
                                            "referencedArticles": [
                                                {
                                                    "bundleId": 23456
                                                }
                                            ]
                                        }
                                    }
                                }
                            }
                        }
                    ]
                },
                "matchingRules": {
                    "$.body.articles": {
                        "min": 0,
                        "match": "type"
                    },
                    "$.body.articles[*].variants.*.bundles.*.referencedArticles": {
                        "min": 0,
[...]

PR #318, opened 2016-09-19
Sample test case for issue #313

cochrane343:
Hi Ronald,

thanks for working on the issue and for your prompt answers! Unfortunately we still can not get our test case working, even after your latest commits. This is why we created a pull request with the test case, so that you can get an exact idea what is not working at the moment.

The `ArticlesTest` generates our pact file. The `ArticlesContractTest` verifies the generated file against the content of `articles.json`. In order to run the test you can simply serve `articles.json` on `http://localhost:8000`, e.g. by executing `python -m SimpleHTTPServer` in the respective folder.

I hope it will help to fix the issue. Thanks a lot in advance!

Best regards,
Alex & Alexandra (pairing)


uglyog:
Cool, hopefully I can get this sorted for you.

Issue #321, opened 2016-09-26
[Reports] Allow dynamic reporters


lordofthejars:
In class https://github.com/DiUS/pact-jvm/blob/master/pact-jvm-provider/src/main/groovy/au/com/dius/pact/provider/reporters/ReporterManager.groovy#L20 you can see that if reporter name is not set an exception is thrown. Why not instead of throwing an exception, we don't try first to treat that value as fully qualified class name and instantiate it as custom reporter?

uglyog:
That's a really good idea.

lordofthejars:
Ok, I will provide a PR soon

uglyog:
Released with version 3.3.1

PR #324, opened 2016-09-27
[issue #321] Allow dynamic reporters providing fully qualified class name

lordofthejars:
Allow dynamic reporters providing fully qualified class name


lordofthejars:
Running CodeNarc locally does not show any violation, and in case of Travis fails with two violations but I cannot see the report

uglyog:
All ok, the codenarc was pretty minor.

Issue #330, opened 2016-10-14
[Enhancements] Some enhancements for Pact-JVM


lordofthejars:
Hello, today I have seen some minor things to improve how users write Pact contracts, I will enumerate here, so we can discuss, the ones we think it would be great to implement, I will open an issue separately and I will provide a PR.

All the enhancements are in `DSLPart` elements:
- [x] implements a `.headers(String... headers)` so instead of having to create a separate `Map`, you can do something like `.headers("content-type", "application/json")`
- [ ] implements a `.numberMatcher()` method like it happens with `stringMatcher`
- [x] implements a `.body(java.util.Consumer<>)` method. Now users can only use JSONObject, Document or String, but if we create a functional method then users wil be able to use lambda to for example transform Jackson POJO to String or JSR 364 JsonObject.
- [x] implements a `toJson` method that allows users to set Json as string using single quotes, so avoiding having to skip over and over again the (") char. https://github.com/arquillian-testing-microservices/game-service/blob/contract-tests/src/test/java/org/arquillian/microservices/gameservice/AgeCheckerConsumer.java#L82

As I said I can provide a PR for each of one.

uglyog:
These are all great ideas.

Just a question on the `.numberMatcher()`. Would the behaviour be to convert the number to a string and then match with a regex?

lordofthejars:
Yes that was my idea

lordofthejars:
I have been thinking about `numberMatch` method and it is quite difficult and I think it will require some deep changes which I don't know if it has sense. Basically if we use the `toString` method to pass form Number to String then the problem is not very big, but I think this approach is a bit useless so we should think about adding a method so users can send how the conversion should be done. The problem comes in stub http server which also needs to know how the conversion should be done to validate what it is received. Currently matchers are stored in a map that contains  the field where is applied and the regexp. This map should be changed to more complex object where you can set also how the ToString thing might happen.

If you agree in this change I can provide the PR as well. Meanwhile I sent https://github.com/DiUS/pact-jvm/pull/331 to review with the other changes.

uglyog:
One option for the number matcher, is to use the number format strings (NumberFormat class) to verify the number's string form matches the format pattern. This is how the date matches work.

lordofthejars:
Hi look I have created this method:

```java
/**
     * Attribute that must match the regular expression
     * @param name attribute name
     * @param regex regular expression
     * @param number example number to use for generated bodies
     */
[...]

PR #331, opened 2016-10-17
Issue #330 Implementing several minimal features in PactDsl

lordofthejars:
Implementing several minimal features in PactDsl.

- implements a `.headers(String... headers)` so instead of having to create a separate `Map`, you can do something like `.headers("content-type", "application/json")`

- implements a `.body(java.util.Consumer<>)` method. Now users can only use JSONObject, Document or String, but if we create a functional method then users wil be able to use lambda to for example transform Jackson POJO to String or JSR 364 JsonObject.

- implements a `toJson` method that allows users to set Json/xml as string using single quotes, so avoiding having to skip over and over again the (") char. https://github.com/arquillian-testing-microservices/game-service/blob/contract-tests/src/test/java/org/arquillian/microservices/gameservice/AgeCheckerConsumer.java#L82


lordofthejars:
One question does this PR also rebased to 3.5 beta too?

uglyog:
I'll merge master into v3-spec branch, which 3.5.x is built off.

PR #346, opened 2016-11-11
An alternative set of helpers for building a PactFragment in Unit Spec mode

alonpeer:
This is an alternative method for building a `PactFragment` when writing specs in Unit Spec mode.
We find this method useful in my team when the provider has many endpoints with which the consumer interacts, in which case the other method of defining the `PactFragment` generates a very unreadable code. Using this method, we can divide our code into smaller pieces and merge them all together in a more readable manner.


alonpeer:
Please ignore this, it's WIP and meant for internal evaluation.

alonpeer:
@uglyog WDYT?

uglyog:
Looks good to me.

PR #350, opened 2016-12-14
Add the option to fetch provider pacts with a specified tag

olga-vasylchenko:
This will allow to verify all the latest consumer pacts with a specified tag for a provider.


uglyog:
Thank you for this, it is very useful and can now be extended to the other build plugins.

PR #352, opened 2016-12-16
Making state comparison case-insensitive

jpava001:
Sometimes consumers use different casing for their provider states which causes problems when the provider tries to run its pact tests because it performs a case-sensitive comparison when determining which method annotated with @State corresponds to the pact.

This problem is magnified when using the Pact Broker since the it changes the casing of the provider state when it renders the pact through the web ui. If a developer writes provider contract tests based on the provider state they see on the Pact Broker, the casing may end up not matching the casing of the actual pact. 

It took me a while to troubleshoot the cause and I think this PR can help some folks.


uglyog:
Thanks for that, you are totally correct, it should be case insensitive.

PR #355, opened 2016-12-22
Improve docs for provider-junit tags

BenSayers:
While using the provider-junit runner we had to look into the implementation to understand how tags work. This update to the docs is capturing our investigation to make things easier for others.


uglyog:
That you, that description was definitely needed.

Issue #356, opened 2016-12-23
Fetching pacts from broker for an unknown provider should fail gracefully


alonpeer:
I'm adding a verification step for a new provider service, which still doesn't have any consumer pacts written for, and so no pacts exist for it on our pact broker (it's not a participant in the broker yet). This means that `http://pact-broker/pacts/provider/new_provider/latest` returns a `404 Not Found`.

This should be considered OK by pact-jvm, because I don't think it makes sense to manually push some provider info to pact-broker just so that it doesn't fail, until a real pact is pushed by some consumer.

At the moment, `HalClient` fails with:

```
DEBUG au.com.dius.pact.provider.broker.HalClient - Fetching: http://pact-broker/pacts/provider/new_provider/latest
au.com.dius.pact.provider.PactVerifierException: Failed to load pacts from pact broker at 'http://pact-broker'
...
Caused by: groovyx.net.http.HttpResponseException: Not Found
	at groovyx.net.http.RESTClient.defaultFailureHandler(RESTClient.java:263)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.codehaus.groovy.reflection.CachedMethod.invoke(CachedMethod.java:93)
	at groovy.lang.MetaMethod.doMethodInvoke(MetaMethod.java:325)
	at groovy.lang.MetaClassImpl.invokeMethod(MetaClassImpl.java:1215)
	at groovy.lang.MetaClassImpl.invokeMethod(MetaClassImpl.java:1082)
	at groovy.lang.MetaClassImpl.invokeMethod(MetaClassImpl.java:1024)
	at groovy.lang.Closure.call(Closure.java:414)
	at groovyx.net.http.HTTPBuilder$1.handleResponse(HTTPBuilder.java:503)
	at org.apache.http.impl.client.CloseableHttpClient.execute(CloseableHttpClient.java:222)
	at org.apache.http.impl.client.CloseableHttpClient.execute(CloseableHttpClient.java:164)
	at groovyx.net.http.HTTPBuilder.doRequest(HTTPBuilder.java:515)
	at groovyx.net.http.RESTClient.get(RESTClient.java:119)
[...]

PR #357, opened 2016-12-23
Gracefully handle requests to pact broker with unknown pacticipants

alonpeer:
This should fix issue #356.


alonpeer:
I'm not sure why checks fail, it seems ok when I run it locally.

uglyog:
Just CodeNarc failures, not a problem, I'll correct those and merge it.

PR #358, opened 2016-12-28
Moved getting provider and consumer names to seperate method

marcodejongh:
For running Pact tests on CI we use system properties for passing in the consumer/providername/tags/etc. Our current implementation touches the private internals of this package, I'm trying to improve it by extending the PactRunner class instead. 

For tags this is easy because I can just override the ```getPactSource``` method but the consumer and provider system props are a bit harder because they are retrieved in the constructor and not a seperate method. To fix this I've moved getting the provider and consumer name to a seperate method which would allow me to override those methods for our runner.

### DIFF
--- a/pact-jvm-provider-junit/src/main/java/au/com/dius/pact/provider/junit/PactRunner.java
+++ b/pact-jvm-provider-junit/src/main/java/au/com/dius/pact/provider/junit/PactRunner.java
@@ -132,7 +132,7 @@ public class PactRunner extends ParentRunner<InteractionRunner> {
         return consumerInfo != null ? consumerInfo.value() : null;
     }
 
-    private String getProviderName(Class<?> clazz) throws InitializationError {
+    protected String getProviderName(Class<?> clazz) throws InitializationError {
         final Provider providerInfo = clazz.getAnnotation(Provider.class);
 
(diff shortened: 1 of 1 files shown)


pact-jvm-provider-junit/src/main/java/au/com/dius/pact/provider/junit/PactRunner.java
```diff
@@ -131,4 +126,19 @@ protected PactLoader getPactSource(final TestClass clazz) throws InitializationE
             throw new InitializationError(e);
         }
     }
+
+    protected String getConsumerName(final Class<?> clazz) {
+        final Consumer consumerInfo = clazz.getAnnotation(Consumer.class);
+        return consumerInfo != null ? consumerInfo.value() : null;
+    }
+
+    private String getProviderName(Class<?> clazz) throws InitializationError {
```
BenSayers: Should this method be protected too?

marcodejongh: Yes it should, I've pushed a fix

marcodejongh:
Closing for now needs more work

PR #374, opened 2017-02-02
Fix logic for default port in case of https url passed to HttpTarget

aniket-21:
Currently if https URL is passed to HttpTarget, the port defaults to 8080 instead of 443


uglyog:
Thanks for that!

PR #388, opened 2017-03-10
Simplifying line separator logic

aniket-21:
This change directly converts stream to string w/o having to read line by line and then concat by System dependent line separator. As a result whatever line separator comes from user data is retained as it is.

Reason for this change is as follows:

I was creating tests for multipart-formdata in which file content is sent. Despite proper line separator ("\n") on linux, my tests were failing. reason being as per the [RFC ](https://www.w3.org/Protocols/rfc2616/rfc2616-sec3.html#sec3.7) 

Extract:
> The message body is itself a protocol element and MUST therefore use only **CRLF** to represent line breaks between body-parts

I was unable to generate requests having CRLF with existing code on Linux machine. This current fix will remove totally the complexity of deciding on line separator and retain whatever is sent as part of body


uglyog:
Do you no longer want to merge this (the PR is closed)?

uglyog:
Oh, I see that this is a duplicate of #393

PR #391, opened 2017-03-17
Update PactBrokerLoader to accept tags from system properties.

Fitzoh:
Currently there is no way to modify the list of tags which are pulled from the broker at runtime.  This commit makes the PactBrokerLoader honor system property expressions in a manner similar to the way it is handled for other parameters. If multiple tags are included in a system parameter they must be comma separated.


Fitzoh:
fixed codenarc stuff

PR #393, opened 2017-03-22
Simplifying line separator logic

aniket-21:
This change directly converts stream to string w/o having to read line by line and then concat by System dependent line separator. As a result whatever line separator comes from user data is retained as it is.

Reason for this change is as follows:

I was creating tests for multipart-formdata in which file content is sent. Despite proper line separator ("\n") on linux, my tests were failing. reason being as per the RFC

Extract:

**The message body is itself a protocol element and MUST therefore use only CRLF to represent line breaks between body-parts**

I was unable to generate requests having CRLF with existing code on Linux machine. This current fix will remove totally the complexity of deciding on line separator and retain whatever is sent as part of body

> Sorry for the last 2 commits, They were not intended to be part of pull request

### DIFF
--- a/pact-jvm-consumer/src/main/java/au/com/dius/pact/consumer/dsl/PactDslJsonArray.java
+++ b/pact-jvm-consumer/src/main/java/au/com/dius/pact/consumer/dsl/PactDslJsonArray.java
@@ -2,6 +2,7 @@ package au.com.dius.pact.consumer.dsl;
 
 import au.com.dius.pact.consumer.InvalidMatcherException;
 import com.mifmif.common.regex.Generex;
+import groovy.json.JsonSlurper;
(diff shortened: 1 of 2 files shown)


alonpeer:
👍 
Would be nice to have a unit test that shows how this fixes the problem you described.

aniket-21:
https://github.com/DiUS/pact-jvm/blob/master/pact-jvm-consumer/src/test/java/au/com/dius/pact/consumer/MimeTypeTest.java#L41

This test actually takes care of the fix. I can explicitly add a CRLF test and LF Test but it actually depends on which machine the test is being executed

alonpeer:
Ah I see. In that case, I guess this is fine as is.

Issue #241, opened 2016-04-06
SSL Support


TheNitek:
I'd like to use pact to test a client implementation which is hard wired to use SSL, but I could not find any hints in the pact-jvm-consumer-junit documentation. Is this possible? If not, I'd like to request this feature.

uglyog:
The current mock server starts up as an HTTP server. It will not be hard to modify it to be able to start up as a HTTPS server. The problem will be what SSL certificate to use. 

We could use a self-signed certificate, but most clients will reject the connection as the certificate does not have a root CA certificate in the clients trust store.

The other option would be for you to provide a certificate that your client will accept.

TheNitek:
In our case both options would be fine. Our client is using a custom key store and trust store anyway, so I could either create test certificates and provide them to the mock server or feed the mock servers certificates to our client.

uglyog:
Getting the mock server to handle HTTPS connections with a self-signed cert was quite easy. If this works for you, I can expose the cert so you can add it to the clients trust store.

TheNitek:
If the .jks file is part of one of the .jar files it should be fine (according to http://stackoverflow.com/questions/344748/how-to-use-a-file-in-a-jar-as-javax-net-ssl-keystore it's quite easy to load a truststore from within a jar) - Otherwise it would not be a problem for me to create my own test trust store with the test certificate, so providing it separately would also work.
Thank you!

uglyog:
It does not use a key store file, it uses certs in PEM format. But the certificates are generated into a temporary folder, so you will be able to access them during the test.

TheNitek:
If I understand correctly there is a new certificate generated on each run? That will work, but I'd prefer to one default certificate in the .jar or being able to provide my own - Most Java SSL things use Java key stores and generating those for every test is quite a pain

uglyog:
Yes, a new cert + private key is generated for each test run. The other option is to provide the cert and private key which will require a little bit more work.

TheNitek:
Ok. Thank you very much for the clarification and the very fast implementation!

uglyog:
Initial version with self-signed certs has been released with versions 3.2.7 and 2.4.9

bbarke:
I have found the need of using a personal keystore for the mock provider that is launched via netty. I'm working on a PR to enable this.

uglyog:
@bbarke I've also started working on a new mock server implementation which I could include using a keystore with.

uglyog:
[...]

PR #420, opened 2017-05-04
Added ability to use a keystore for the pact mock server #241

bbarke:
Added ability to use a keystore

Updated JUnit docs and included tests.


uglyog:
That's awesome, thanks for the PR.

PR #425, opened 2017-05-11
#241 Added ability to use trust store in pact-jvm-server

bbarke:


uglyog:
Awesome, thanks for the PR!

PR #426, opened 2017-05-12
Added method to PactFolderLoader to provide files along with pacts to…

raceconditions:
… fix the ConsumerInfo instantiation.

We found that AmqpTarget does not handle @PactFolder annotation well as something deep into the framework chokes when ConsumerInfo is missing the File version of the pact.  This felt somewhat hacky to do, but a new method in PactFolderLoader to return both the pact and the file, then creation of ConsumerInfo in AmqpTarget using the Consumer name and the pact File fixes the issue.


uglyog:
Thanks for the PR.

I agree it is all a bit hacky. I have tried to implement pushing verification results back to the pact broker, but how pact files are sourced and provided to the verifier needs to be re-designed.

PR #427, opened 2017-05-12
Added RestPactRunner and MessagePactRunner

raceconditions:
When a microservice is a provider of both a REST API and Messages, the PactRunner cannot differentiate between HTTP/AMQP pacts and thus attempts to play both types against the respective test type.  This provides differentiation via inheritance and filters to only play certain Pact types against a provider.

We didn't update docs yet, so feel free to either reject or I can follow up with a subsequent PR with doc updates.  The change is not breaking.


uglyog:
This looks similar to #426 but I have merged it in anyway, just in case.

raceconditions:
Yeah sorry about that, my intent was to separate the PRs but I must have given you #426 with both.

PR #435, opened 2017-05-18
Spring module

raceconditions:
Support for Spring tested to work with Spring 3 and Spring 4.  Hard dependency on pact-jvm-provider-junit.  Provides a new Target which can be configured for Spring controllers using Spring-Test MockMvc.


uglyog:
Awesome, I actually know someone who was asking for this functionality last year.

Issue #441, opened 2017-05-23
Problem with eachKeyLike generated matchingRules


lucaong:
Hi,
first of all, thanks for the great work :)

I encountered a problem with `eachKeyLike(String exampleKey, PactDslJsonRootValue value)`. I have a JSON body like:

```java
.body(new PactDslJsonBody()
  .object("foo")
    .eachKeyLike("123", PactDslJsonRootValue.numberType(42))
  .closeObject()
)
```

The generated body is correct:

```javascript
"body": {
  "foo": {
    "123": 42
  }
}
```

However, the generated rules are not:

```javascript
"matchingRules": {
  "$.body.*": {
    "match": "type"
  }
}
```

While I would expect:

```javascript
"matchingRules": {
  "$.body.foo.*": {
    "match": "type"
  }
}
```

Note that the overload of `eachKeyLike` taking only one argument works as expected, but I cannot use it, as I need to match a root value (a number in my case). Am I doing it wrong or is this a bug?

I suspect the problem is around here: https://github.com/DiUS/pact-jvm/blob/393429ecab95f55f2b55def5f6bed1b85355faac/pact-jvm-consumer/src/main/java/au/com/dius/pact/consumer/dsl/PactDslJsonBody.java#L972 probably in the rule generation logic. I could dedicate some time to a pull request if it is confirmed to be a bug.

lucaong:
I sent the following pull request: https://github.com/DiUS/pact-jvm/pull/444

lucaong:
Thanks @uglyog for taking time to review and merge the PR :)

Do you know when this fix can get to the Maven repository?

lucaong:
oh, I see it's out now, in release `3.5.0-rc.1`. Thanks a lot!

PR #444, opened 2017-05-24
fix the matchingRules generated by eachKeyLike with PactDslJsonRootValue

lucaong:
when `eachKeyLike(String exampleKey, PactDslJsonRootValue value)` is called, it generates matchingRules that are mistakenly not namespaced under the rootPath (see issue https://github.com/DiUS/pact-jvm/issues/441 for an example). This commit should fix it, and adds a test for it.

Please review because I am not very familiar with the codebase ;)


uglyog:
Looks good, thanks for the PR

PR #446, opened 2017-05-30
Marcpa00 readme fixes

marcpa00:
If one follows instructions in README.md of pact-jvm-consumer-groovy, the code does not compile because PactVerificationResult is not found in the version mentionned (3.3.8).  
These commits tell to use 3.4.0 and include a complete test class (with imports and class declaration).


uglyog:
Thank you for the PR.

PR #450, opened 2017-06-01
Boolean to control pactVerify.dependsOn a provider task

marcpa00:
Every provider task defined becomes a dependency on `pactVerify`.  

Now suppose that your provider task defines start/stop provider, but the new provider version fails to verify : would be nice to be able to start manually the provider (under a debugger, for example) and simply re-execute the pacts defined.  Right now, to do this, one would have to change the build.gradle file to put the definition of `startProviderTask` and `terminateProviderTask` in comments, then re-run `pactVerify` (and remember not to commit this change... otherwise the CI build previously assuming the provider is automatically started by `pactVerify` will fail).

With this PR, boolean `isDependencyForPactVerify` is added to provider definition and the dependency chaining between `pactVerify` and the new provider task is done only if this is true (the default) and skipped if false.  This way, one can duplicate a provider definition, give it another name (lets say "dev"), remove start/stop provider properties, set `isDependencyForPactVerify` to false to have best of both worlds : `gradle pactVerify` will start the provider, verify, kill the provider as usual, and `gradle pactVerify_dev` can be used after having manually started the provider to run the pacts against it.

### DIFF
--- a/pact-jvm-provider-gradle/README.md
+++ b/pact-jvm-provider-gradle/README.md
@@ -164,6 +164,37 @@ pact {
 Following typical Gradle behaviour, you can set the provider task properties to the actual tasks, or to the task names
 as a string (for the case when they haven't been defined yet).
 
+## Preventing the chaining of provider verify task to `pactVerify` [version 3.4.1+]
+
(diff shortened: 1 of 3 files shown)


uglyog:
This is a great idea, thanks for the PR.

PR #458, opened 2017-06-09
Pact pr

siddhuwarrier:
This PR creates a filtered pact runner that allows you to filter interactions in a pact specified by provider state so that you only run interactions within a pact that have specific provider states, mirroring the behaviour provided by the Groovy and Maven plugins.

---
This change is [Reviewable](https://reviewable.io/reviews/dius/pact-jvm/458)


uglyog:
Thanks for the PR, that looks useful.

siddhuwarrier:
Thank you for merging it! 👍

PR #461, opened 2017-06-14
Allow an empty string to be specified when using the Groovy string matcher

jon-ruckwood:
When using the string matcher with a non-null value you expect to see that example value in the Pact file. This is not the case with an empty string, a generated value would be present in the Pact instead. My pull request addresses this situation.

Before my change `Matchers.string()` was equivalent to `Matchers.string('')`. Now it is possible to use an empty string as an example value.

This change also makes the matcher's behaviour consistent with the equivalent Java matcher ([`PactDslJsonBody#stringType(...)`][0]).

[0]: https://github.com/DiUS/pact-jvm/blob/master/pact-jvm-consumer/src/main/java/au/com/dius/pact/consumer/dsl/PactDslJsonBody.java#L130-L139


uglyog:
Thanks for the PR

PR #465, opened 2017-06-19
Make message pact metadata available from MessagePactProviderRule

andreschaffer:
Hi,
I see there is a way of defining metadata in the MessagePact (that is, the return from the @Pact annotated method on the consumer side), but there is now way of retrieving it from the MessagePactProviderRule.
I added code and tests for retrieving it.
If that is something you find needed, PACT then should probably support verifying that metadata in the provider side as well. I can try and find the time for another PR regarding that - no promises :)


uglyog:
Thanks for the PR

PR #468, opened 2017-06-22
Add a Groovy boolean matcher

jon-ruckwood:
I've added a boolean type matcher. I checked the spec and I think I did correctly?

Ideally I would've called it `boolean` but that's reserved. `bool` seemed like a somewhat reasonable alternative, but it's not a term used much in the JVM-world (I have no strong feelings about the name being changed).


uglyog:
Oh, wow, I'm astounded that that was missing. Thanks for the PR.

PR #470, opened 2017-06-23
Allow setting of a custom MockMvc for provider states

mouserd:
Hi,

I've made this change to the **pact-jvm-provider-spring** library to allow a provider state to provide it's own MockMvc that will be used during the test interaction. This is useful in situations where more control is required over the Spring test context.

The example test that I've added to exercise the new behaviour demonstrates a scenario where a Spring controller needs a custom argument resolver to process a "rich" request parameter.

Let me know if you have any questions or comments.

Thanks,
Dave


uglyog:
Awesome, Thanks for the PR!

PR #481, opened 2017-07-10
Fix bodyWithSingleQuotes with ContentType param

jeffreymyers:
Call to bodyWithSingleQuotes(String body, ContentType mimeType) did not behave as intended - delegated to wrong method.

Add documentation for mimeType method params.


uglyog:
Good catch, Thanks!

PR #484, opened 2017-07-13
Updating URL matching regex to apply to file URLs

jeffreymyers:
Fixes an issue matching on file:/ URLs. Before the change, this usage of the junit `@PactUrl` caused PactReader to throw confusing exception when it tried to interpret the URL value as a json directly on line 185. This change resolved the issue.

e.g.
`@PactUrl(urls = {"file:/path/to/pact.json"})`


uglyog:
Wow, one character has so much power!

Issue #455, opened 2017-06-08
failIfNoPactsFound = false only changes exception thrown


Nalha:
Running my provider tests with the annotations:
```
@Provider("test_provider")
@PactBroker(failIfNoPactsFound = false, host = "localhost", port = "8080")
@RunWith(RestPactRunner.class)
```

throws this exception:
```
java.lang.Exception: Did not find any pact files for provider test_provider

	at org.junit.runners.model.InitializationError.<init>(InitializationError.java:38)
	at au.com.dius.pact.provider.junit.PactRunner.<init>(PactRunner.java:81)
```

While running it with failIfNoPactsFound = true throws this instead
```
u.com.dius.pact.provider.junit.loader.NoPactsFoundException: No consumer pacts were found for provider 'test_provider' and tag 'null'. (URL [])

	at au.com.dius.pact.provider.junit.loader.PactBrokerLoader.loadPactsForProvider(PactBrokerLoader.java:83)
```
Is there a way to disable the runner when pacts are missing or is this a broken feature?

uglyog:
This is a broken feature.

stemlaur:
Hello @uglyog, I encounter the same issue for the version 3.5.2, do you know what is the last version where this property `failIfNoPactsFound` feature was working ?

uglyog:
There is a regression introduced in 2.4.1, so 2.4.0 should have the old behaviour.

This issue is on conflict with #188, so it needs to be fixed for both uses.

uglyog:
I have deprecated the `failIfNoPactsFound` attribute and added a new `@IgnoreNoPactsToVerify ` annotation that can be added to the test class to control this behaviour.

uglyog:
Version 3.5.3 has been released

Issue #188, opened 2015-11-24
No feedback when using an invalid @Provider name in pact-jvm-provider-junit


stefan-lz:
When using a @Provider with a provider name that does not exist, eg:

@RunWith(PactRunner.class)
@Provider("Provider Service WithTypo")
@PactFolder("xyz")
class PactTest {
..
..
}

No tests are run, and no error is reported and I get:

    Process finished with exit code 0

It's not finding any interactions to test, but I would expect an error or something.

uglyog:
The other provider implementations display a warning when this is the case.

uglyog:
Looking through the implementation, it will probably have to be an initialization error.

stefan-lz:
Awesome - thanks for the quick fix.

knok16:
Originally this piece of code was developed without such check intentionally because sometimes absence of pacts is legitimate situation. For example, we have a lot of microservices, use pact-broker, and code for running contract tests against provider are added when service is developed, and at this point of time service have not any consumer, i.e. have no any pacts in pact broker.

stefan-lz:
@knok16, I raised this because the general idea is to not write any provider code until you have a pact from the consumer to test against it. I guess nothing is technically failing, so I'd be happy with a loud enough warning in the console too.

PR #488, opened 2017-07-17
Fix issue 455

stemlaur:
Commit 468429677703620469a7fc13684882db70341edd introduced a feature
regression (see https://github.com/DiUS/pact-jvm/issues/455): it is not
possible anymore to write provider tests before having developed any
consumer expectations.
@PactRunner should accept having no pacts to execute for that matter.


uglyog:
This was changed for issue #188, so we can't just roll the changes back. I need to fix it for all use cases.

uglyog:
I'm going to leave the default to fail if there are no pacts to verify, and then add a annotation to control this behaviour. I will close this PR and then make the changes against #455

Issue #415, opened 2017-04-28
Using @PactBroker, provide tag with a system property


EstherAF:
Hi,

I'm using @PactBroker, with host and port values taken from system properties, like follows:
`@PactBroker(host = "${pact.broker.host}", port = "${pact.broker.port}", tags = "latest")`
and was working great.

Now I've tried to do the same with the 'tags' attribute, like follows:
`@PactBroker(host = "${pact.broker.host}", port = "${pact.broker.port}", tags = "${pact.provider.tag}")`
but throws this error:

> Caused by: java.lang.RuntimeException: Could not resolve property "${pact.provider.tag}" in the system properties or environment variables and no default value is supplied
        at au.com.dius.pact.provider.junit.sysprops.SystemPropertyResolver.resolveValue(SystemPropertyResolver.java:18)
        at au.com.dius.pact.provider.junit.sysprops.PactRunnerTagListExpressionParser.substituteIfExpression(PactRunnerTagListExpressionParser.java:27)
        at au.com.dius.pact.provider.junit.sysprops.PactRunnerTagListExpressionParser.lambda$parseTagListExpressions$0(PactRunnerTagListExpressionParser.java:21)
        at java.util.stream.ReferencePipeline$7$1.accept(ReferencePipeline.java:267)
        at java.util.Spliterators$ArraySpliterator.forEachRemaining(Spliterators.java:948)
        at java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:481)
        at java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:471)
        at java.util.stream.ReduceOps$ReduceOp.evaluateSequential(ReduceOps.java:708)
        at java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234)
        at java.util.stream.ReferencePipeline.collect(ReferencePipeline.java:499)
        at au.com.dius.pact.provider.junit.sysprops.PactRunnerTagListExpressionParser.parseTagListExpressions(PactRunnerTagListExpressionParser.java:22)
        at au.com.dius.pact.provider.junit.sysprops.PactRunnerTagListExpressionParser.parseTagListExpressions(PactRunnerTagListExpressionParser.java:16)
        at au.com.dius.pact.provider.junit.loader.PactBrokerLoader.<init>(PactBrokerLoader.java:44)
[...]

PR #493, opened 2017-07-22
Fix placeholders resolution in "tags" field of @PactBroker annotation

dmikhievich:
Fix of bug #415 including unit tests update


uglyog:
Thanks, I have just been impacted by this today!

PR #510, opened 2017-08-05
update `PactBroker` to allow setting properties via system properties by default

Fitzoh:
For some reason master wasn't building on my machine.
I have more documenting to do but wanted to see what travis had to say.


cah-andrewfitzgerald:
Updated with documentation

Fitzoh:
Updated with documentation

Fitzoh:
This works, but feels a little bit janky. Any thoughts @uglyog ?

uglyog:
Looks good to me. I'll change the default authentication scheme to `basic`, as `none` does not make sense, now that I think of it.

PR #519, opened 2017-08-15
Use UTF-8 as a default charset in MockServer

Windowsfreak:
Use UTF-8 instead of ISO-8859-1 to parse bodies if no charset has been provided from the client.
If a client sends a request using Content-Type: application/json, the behaviour should be identical to a client requesting using Content-Typre: application/json, Charset=UTF-8.


Windowsfreak:
Another fix for this problem would be to add another Constructer parameters for classes such as PactProviderRuleMk2, that allow setting the default Charset, that is then saved to the MockProviderConfig being instantiated. This will be sent to the MockServer's constructor, which may then use a different default encoding when reading bodies.

PR #520, opened 2017-08-17
Enable access to test clazz for custom annotations

bbarke:
Allow access to test clazz for custom annotations.


uglyog:
There is a `getTestClass()` method on the parent runner, which you can use to get the class under test (it has a `getJavaClass()` to get the actual Java class.

bbarke:
You are right, thank you!

Issue #445, opened 2017-05-26
Scala 2.12 version


ajhcpt:
Hi, when will there be a release compatible with Scala 2.12?

Regards,
Andrew

SanCoder-Q:
I have the same question.

jcreixell:
+1

BenjaminDebeerst:
I have been looking into this, and there is 4 dependencies that need an upgrade to be able to provide Scala 2.12 support as their current version don't have 2.12 support. This is:

* `com.typesafe.scala-logging:scala-logging` is currently at 3.4.0 and needs update to [at least 3.5.0](https://mvnrepository.com/artifact/com.typesafe.scala-logging/scala-logging_2.12)
* `net.databinder:unfiltered-netty-server` [currently has no scala 2.12 support](https://mvnrepository.com/search?q=net.databinder&p=1), but the project is active and 2.12 support is [in the making](https://github.com/unfiltered/unfiltered/blame/0.10.x/project/common.scala#L15)
* `net.databinder.dispatch:dispatch-core` is currently at 0.11.3 and needs update to [at least 0.12.0](https://mvnrepository.com/artifact/net.databinder.dispatch/dispatch-core_2.12)
* `io.gatling:jsonpath` is currently at 0.6.4 and needs update to [at least 0.6.8](https://mvnrepository.com/artifact/io.gatling/jsonpath_2.12)

I'll be working on PRs to update dependencies for those projects that do have Scala 2.12 support already.

BenjaminDebeerst:
Contrary to what I wrote in my first comment, `unfiltered-netty-server` does have support for Scala 2.12 since version 0.9.0, and that version is already out. What I missed (and the reason why I didn't find it), is that the project group ID has been [renamed to `ws.unfiltered`](https://github.com/unfiltered/unfiltered/commit/91b81b318f0355c774f5c4130fb34b10f2f9fa1b). Versions 0.9.0 and 0.9.1 of the project have been released [under that new org name](https://mvnrepository.com/artifact/ws.unfiltered).

uglyog:
Scala 2.12 version 3.5.4-rc.0 has been released

kwark:
It looks like this is not working yet: I am getting following resolution problems

[warn] 	Note: Unresolved dependencies path:
[warn] 		au.com.dius:pact-jvm-model:3.5.4-rc.0
[...]

PR #524, opened 2017-08-22
Update io.gatling:jsonpath dependency to 0.6.9

BenjaminDebeerst:
This is one of several necessary version upgrades to be able to provide Scala 2.12 support. (See #445)

[Since version 0.6.7](https://github.com/gatling/jsonpath/compare/bbced508133...404eee2e95#diff-ae7d7c0ef89d66c7f95b5f359b182c90R52), Gatling's Json parser allows Json field names starting with digits, which is why the corresponding tests are updated alongside the version upgrade.


uglyog:
I've merged your PR into the scala-2.12 branch.

PR #525, opened 2017-08-22
Upgrade dependencies for Scala 2.12 support

BenjaminDebeerst:
Prepare scala 2.12 support: Update the scala-logging and `dispatch-core` dependencies to versions that provide Scala 2.12 support. See #445.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -79,7 +79,7 @@ subprojects {
       if (project.hasProperty('scalaVersion')) {
           compile "org.scala-lang:scala-library:${project.scalaFullVersion}",
             "org.scala-lang:scala-compiler:${project.scalaFullVersion}",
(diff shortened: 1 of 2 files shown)


uglyog:
I've merged this PR into the scala-2.12 branch

BenjaminDebeerst:
Great, thank you.

I didn't realize a Scala 2.12 branch existed. There has not been much work on that branch over the past year though, can you elaborate on the plans regarding 2.12 support? Do you see other issues other than updating dependencies that need to be resolved to get this into the main development line?  

At SoundCloud, we are very interested in that and would like to contribute to this project to make it happen.

(#445 might be the better place to discuss these questions.)

uglyog:
I created the branch when the first milestone of 2.12 was released, but there were too many dependencies that had no 2.12 version, so it got put on the back burner. It is probably a good time to move the mainline version to 2.12.

Issue #546, opened 2017-09-21
The array matcher: eachLike should require at least one example


ptornhult:
If the eachLike matcher specifies Zero items the service provider implementing the Pact might only specify an empty array which would pass all tests. Once deployed, your provider could return an item that doesn't look like the one you expected, and then you have a bug (which you may or may not notice).

Right now the JVM version and JS version of eachLike behave differently in this regard. They should behave the same I would think/hope. It was a good catch by @bethesque to not allow this to slip into the JS version of Pact, but it should be fixed in the JVM version then. I can make a PR for this if you like?

Read the comments here for more info: https://github.com/pact-foundation/pact-js/pull/95

bethesque:
Perhaps, in the interests of maintaining backwards compatibility for the jvm impl, we could print an obnoxious warning when a 0 length array is specified?

For the Rust impl, I'm pretty keen on not allowing 0 length.

uglyog:
I prefer this to be an error, in the same way as giving an example value that does not match the regex causes an error.

bethesque:
I'd prefer an error too, I just wasn't sure if you'd be cool with making it backwards incompatible.

PR #552, opened 2017-09-28
Print error when eachLike matcher is used with 0 examples

ptornhult:
Using the eachLike matcher with zero examples may be unsafe. But since this is already in use we shouldn't break backward compatibility. In the JS implementation this isn't even allowed, but in the JWM implementation it was suggested to only print an obnoxious message in case eachLike is called with Zero examples when generating pacts.

See: https://github.com/DiUS/pact-jvm/issues/546


uglyog:
I actually prefer to make this an error, same as when using a regex and giving an example string that does not match the regex.

uglyog:
I've merged this, but I'm going to convert it to an error

Issue #567, opened 2017-10-21
Potential issue regarding encoded query parameter matching


remen:
Hi!

I'm having some trouble with url encoding issues in query parameters using `au.com.dius:pact-jvm-consumer-junit_2.12:3.5.8`. Unless I am mistaken, it seems that the mock server double-decodes the query parameter value unnecessarily which makes incoming requests not match correctly.

As an example, I would expect the following test to succeed:

```java
import au.com.dius.pact.consumer.ConsumerPactBuilder;
import au.com.dius.pact.consumer.PactVerificationResult;
import au.com.dius.pact.model.MockProviderConfig;
import au.com.dius.pact.model.RequestResponsePact;
import org.apache.http.client.methods.CloseableHttpResponse;
import org.apache.http.client.methods.HttpGet;
import org.apache.http.client.utils.URIBuilder;
import org.apache.http.impl.client.CloseableHttpClient;
import org.apache.http.impl.client.HttpClients;
import org.junit.Test;

import java.net.URI;

import static au.com.dius.pact.consumer.ConsumerPactRunnerKt.runConsumerTest;
import static org.junit.Assert.assertEquals;

public class EncodingTest {
    @Test
    public void test_pact_query_encoding() throws Exception {
        RequestResponsePact pact = ConsumerPactBuilder
            .consumer("Some Consumer")
            .hasPactWith("Some Provider")
            .uponReceiving("a request to say Hello")
            .path("/hello")
            .method("GET")
            .matchQuery("q", ".+", "http://example.com?q=my%20search&u=123")
            .willRespondWith()
            .status(200)
            .toPact();
        CloseableHttpClient httpClient = HttpClients.createDefault();
        MockProviderConfig config = MockProviderConfig.createDefault();
        PactVerificationResult result = runConsumerTest(pact, config, mockServer -> {
            URI uri = new URIBuilder(mockServer.getUrl() + "/hello")
                .addParameter("q", "http://example.com?q=my%20search&u=123")
                .build();

            // A superfluous test to check that URIBuilder is doing the right thing here.
            assertEquals(uri.toString(), mockServer.getUrl() + "/hello?q=http%3A%2F%2Fexample.com%3Fq%3Dmy%2520search%26u%3D123");

            try (CloseableHttpResponse response = httpClient.execute(new HttpGet(uri))) {
                assertEquals(response.getStatusLine().getStatusCode(), 200);
            }
        });

        if (result instanceof PactVerificationResult.Error) {
[...]

PR #570, opened 2017-10-24
Fix double-decoding of query parameters

remen:
As mentioned in issue #567, the MockHttpServer was double-decoding query parameters. This PR fixes this, adds some more tests and fixes an existing test that was double-encoding parameters before sending.

Furthermore, to keep the existing API, this PR does not change the behavior of the `.query(...)` DSL method which expects a *unencoded* query string. Instead it adds a `.encodedQuery(...)` method that is expected to be encoded. This allows for it to contain `&` and `=` characters.

Fixes: #567


uglyog:
Thanks for the PR!

PR #577, opened 2017-10-31
Add matchQuery with List as example

### LINKED ISSUE #509: Support for one query parameter multiple times
The service I am calling takes a query parameter named "nodeId". This parameter can have multiple values, so a request URL can look like this:

    http://localhost/path?nodeId=8&nodeId=20&nodeId=999

I tried to build the fragment like this:

            .matchQuery("nodeId", "^[0-9]{1,3}$", "8")
            .matchQuery("nodeId", "^[0-9]{1,3}$", "20")
            .matchQuery("nodeId", "^[0-9]{1,3}$", "999")

However I get the following error when running the consumer pact test:

    QueryMismatch(nodeId,List(999),List(8, 20, 999),Some(Expected query parameter nodeId with 1 values but received 3 values),$.query.nodeId)

While not quite that normal, HTTP allows you to provide the same query parameter multiple times, allowing a single service call to respond with multiple sets of data.

Is this supported and if not, please consider this a feature request :-)

Mordavolt:
As discussed in #509 pact-jvm should be able to allow lists of examples for query param definitions. This commit adds no new functionality, just enables the API to use the existing one. This method will create requests that look like this: /path?param=A&param=B

Also fixed some formatting

Fix #509

### DIFF
--- a/pact-jvm-consumer/src/main/java/au/com/dius/pact/consumer/dsl/PactDslRequestWithPath.java
+++ b/pact-jvm-consumer/src/main/java/au/com/dius/pact/consumer/dsl/PactDslRequestWithPath.java
@@ -340,24 +340,39 @@ public class PactDslRequestWithPath {
         return new PactDslResponse(consumerPactBuilder, this);
     }
 
-  /**
-  * Match a query parameter with a regex. A random query parameter value will be generated from the regex.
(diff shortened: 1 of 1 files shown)


uglyog:
Thanks for the PR

PR #579, opened 2017-11-01
Update readme links

shashidesai:
* Update Pact Broker and Pact specification links
ps: pact users are getting redirected to a inactive repo for the broker, so thought of updating the readme.


uglyog:
Thanks for the PR

PR #596, opened 2017-12-01
add support for asyncResult()

lukasGemela:
All our controllers are written in async way, returning CompletableFutures with the response. 

Pact-jvm does not support that feature and tries to evaluate and mach HTTP response immediately (when its actually not loaded), failing on assertion because it's trying to match pact spec against default response values. 

following https://stackoverflow.com/questions/17711615/testing-spring-asyncresult-and-jsonpath-together, this patch enables testing of controllers with completable futures


uglyog:
Thanks for the PR.

PR #607, opened 2017-12-21
Fixes error when removing path ending with slash(/)

lordofthejars:


uglyog:
@lordofthejars this PR is marked as closed, do you still need this to be merged?

lordofthejars:
Well it is merged,and fixed an important bug I guess. I managed to fix in Algeron part but I think fixing in upstream is important as well.

Issue #605, opened 2017-12-18
java8 dsl does not properly manage min/maxArrayLike


ohubaut:
The Java8 dsl simply delegate method such as `minArrayLike` to the underlying object, which will internally create a new array. 
But, as there is no `closeArray` method on LambdaDslObject, the array remain unclosed.
Moreover, as the newly created array is not exposed, it is not possible to provide any specification for the given array.
The solution would be to use a similar mechanism as for `array` and `object` method and force a last argument to be the Consumer of that array. Some convenience method without the Consumer could also be provided, which would properly close the internal array immediately.

eatdrinksleepcode:
I have just run into this as well. Do I understand correctly that this means that there is no way using the Java8 DSL to specify an array with an unknown number of items? Is the only choice to fall back to the standard java DSL?

ohubaut:
I have written a pull request that should fix the issue. Feel free to review it (works on my machine 😁) #611

PR #611, opened 2018-01-04
Fix for #605

ohubaut:
This fix allow the LambdaDSL to generate Pact spec that are compliant with those generated with the default JVM Pact DSL.

It is a bit rough on the edges (some code duplication) and break the API (by forcing the use of a consumer), but I do consider it as a minor trade-off in order to have a more fluent, less error-prone way to write Pact Specs

### DIFF
--- a/pact-jvm-consumer-java8/src/main/java/io/pactfoundation/consumer/dsl/LambdaDslJsonArray.java
+++ b/pact-jvm-consumer-java8/src/main/java/io/pactfoundation/consumer/dsl/LambdaDslJsonArray.java
@@ -223,8 +223,11 @@ public class LambdaDslJsonArray {
     /**
      * Element that is an array where each item must match the following example
      */
-    public LambdaDslJsonArray eachLike() {
-        pactArray.eachLike();
+    public LambdaDslJsonArray eachLike(Consumer<LambdaDslJsonBody> nestedObject) {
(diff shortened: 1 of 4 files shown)


eatdrinksleepcode:
> It is a bit rough on the edges (some code duplication) and break the API (by forcing the use of a consumer), but I do consider it as a minor trade-off in order to have a more fluent, less error-prone way to write Pact Specs

I would propose that the current overloads should be maintained as-is, but marked as deprecated in favor of the new ones. In a future release, the current overloads could be removed.

Also, there is a lot of cosmetic change in the tests that would be better to either revert or to pull out into a separate PR, to make understanding this PR easier.

Other, than those two comments, looks good to me.

ohubaut:
> I would propose that the current overloads should be maintained as-is, but marked as deprecated in favor of the new ones. In a future release, the current overloads could be removed.

That statement seems odd to me. The current overloads does not generate correct Pacts specifications. And once they are invoked, the rest of the contract will be corrupted.
At best, I could let them and automatically close the underlying opened arrays, but it would still be extremely confusing for the users as it would give the impression they could use them as in the classic DSL version, while it is not the case.

uglyog:
I also don't see the benefit of the current overloads being maintained as-is, as they are currently not functional without passing a consumer in to define the correct expectation on the contents of the collection.

PR #612, opened 2018-01-04
Fix missing generation of array examples with `eachArray*` methods

ohubaut:
Those methods were not properly generating the array examples to be used by
the Pact consumers, resulting in indirect violation of the contract and/or
incorrect handling of the data by the consumer.

The issue was probably never caught as it seems there is no test case for
example array in the spec except for the trivial "at least one element",
which is always satisfied.


uglyog:
Thanks for the PR

Issue #614, opened 2018-01-08
Some provider names generate gradle 5.0 deprecation warnings


TimothyJones:
Provider names become task names, which means that there are some deprecation warnings generated in recent gradle versions.  For example:

```
pact {
    serviceProviders {
        'Some Provider - With Forbidden Characters' {
     .....
```
generates the following warning:

```
The task name 'pactVerify_Some Provider - With Forbidden Characters' contains at least one of the following characters: [ , /, \, :, <, >, ", ?, *, |]. This has been deprecated and is scheduled to be removed in Gradle 5.0.
```

We should ensure that the task names don't contain those characters - I'll generate a PR for this.

PR #615, opened 2018-01-08
Future proof task names for gradle 5.0

TimothyJones:
This is a fix for the cause of #614.

Gradle's `NameValidator` class exposes an `asValidName()` method which converts names to valid names. However, currently it doesn't do anything (the [comments](https://github.com/gradle/gradle/blob/master/subprojects/core/src/main/java/org/gradle/util/NameValidator.java#L48) say that it will convert task names once the deprecated task names are actually forbidden).

This means that this PR doesn't fix the deprecation warning, but should fix the problem when the deprecation actually happens.

Possible improvements to this PR:

* Suppress the warning
* If that is not possible, generate an additional warning informing users that they can ignore the warning

Alternatives:

* Provide our own implementation of `asValidName` to suppress the warning. I don't like this, as it means that we're not robust to gradle's definition of a valid name changing.

### DIFF
--- a/pact-jvm-provider-gradle/build.gradle
+++ b/pact-jvm-provider-gradle/build.gradle
@@ -7,6 +7,9 @@ apply plugin: 'maven-publish'
 
 dependencies {
   compile project(":pact-jvm-provider_${project.scalaVersion}")
+  testCompile 'org.powermock:powermock-module-junit4:1.7.3'
+  testCompile 'org.powermock:powermock-api-mockito2:1.7.3'
+  testCompile 'org.mockito:mockito-core:2.8.0'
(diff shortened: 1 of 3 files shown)


TimothyJones:
I'm not sure why these builds are failing, but it seems unrelated to the change?

uglyog:
The build failure was due to intermittent network issues: 

> Connect to jcenter.bintray.com:443 [jcenter.bintray.com/108.168.243.150] failed: connect timed out

TimothyJones:
@uglyog I'm happy to not require new test dependencies, but I wasn't sure how to test the static method dependency otherwise. Happy to take guidance on that.

I've changed the code to try to load the class, and to have default behaviour if that fails.

Issue #618, opened 2018-01-10
Release pact-jvm-provider-sbt plugin for version 3.5.x


jcreixell:
The latest released version of `pact-jvm-provider-sbt` is from the 2.x branch, which is incompatible with pact specification 3.x. This causes problems when verifying pacts generated by `pact-jvm-consumer-specs2 ` 3.5.x (which is required for using Scala 2.12) and made available through a broker.

PR #621, opened 2018-01-15
Upgrade provider sbt plugin to 3.x, Scala 2.12, SBT 1.x

jcreixell:
- Updates `pact-jvm-provider-sbt` plugin to support SBT 1.x, Scala
  2.12.
  - Migrates to [auto plugin](http://www.scala-sbt.org/1.0/docs/Plugins.html) API.
  - See https://github.com/DiUS/pact-jvm/issues/618

### DIFF
--- a/pact-jvm-provider-sbt/build.sbt
+++ b/pact-jvm-provider-sbt/build.sbt
@@ -1,9 +1,15 @@
 sbtPlugin := true
 isSnapshot := true
 
+scalaVersion := "2.12.4"
+organization := "au.com.dius"
+
+val currentVersion = "3.5.11"
+version := currentVersion
+
 libraryDependencies ++= Seq(
-  "au.com.dius" %% "pact-jvm-provider-sbtsupport" % Common.version,
-  "ch.qos.logback" % "logback-classic" % "1.1.7"
+  "au.com.dius" %% "pact-jvm-provider-scalasupport" % currentVersion,
+  "ch.qos.logback" % "logback-classic" % "1.2.3"
 )
 
 scalacOptions ++= Seq("-deprecation", "-feature")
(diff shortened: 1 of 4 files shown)


pact-jvm-provider-sbt/src/main/scala/au/com/dius/pact/provider/sbt/SbtProviderPlugin.scala
```diff
@@ -12,27 +12,30 @@ import sbt._
...
-      }
-
-      if (providers.value.isEmpty)
-        sys.error("No providers have been defined. Configure them by setting the providers build setting.")
-      else
-        Verification.verify(providers.value)
-    })
+object SbtProviderPlugin extends AutoPlugin {
+  object autoImport {
+    lazy val pactProviders = SettingKey[Seq[ProviderConfig]]("providers", "Providers to verify")
```
jcreixell: note: `providers` is renamed to `pactProviders`, and `config` to `pactProvidersConfig`. The names are only suggestions, but since the AutoPlugin brings all of that in scope automatically, names need to be more specific. Unfortunately this breaks the old API.

pact-jvm-provider-sbt/build.sbt
```diff
@@ -1,9 +1,15 @@
...
 isSnapshot := true
 
+scalaVersion := "2.12.4"
+organization := "au.com.dius"
+
+val currentVersion = "3.5.11"
+version := currentVersion
+
 libraryDependencies ++= Seq(
-  "au.com.dius" %% "pact-jvm-provider-sbtsupport" % Common.version,
```
jcreixell: I think `Common.version` was broken after the removal of `project/Build.scala`. For now I am setting the project version here, but it needs to be manually updated on every release.

uglyog:
Thanks for the PR.

Ideally this plugin should be moved to it's own repo, which I might do at some point.

PR #622, opened 2018-01-16
Document new API for provider SBT plugin

jcreixell:
Documents recent changes to the plugin setting keys.

See #621


jcreixell:
in addition, this could be a good time to remove the deprecated functionality (`PactJvmPlugin.scala`) since this will be a major version bump.

Issue #439, opened 2017-05-22
Reduce maven transient dependecie


Thadir:
The dependency list of the pact-jvm-provider and consumer for junit, are a bit big. And can cause a lot of conflicts (think about the sl4j dependence thats cross project from pact-jvm-model to pact-jvm-matcher etc).

We should analyse and reduce the needed packages (and if possible reduce the amount of nested dependencies (but this could be a scala thin).

tmszdmsk:
I researched topic of CDC today and, to be honest, I am not sure if I should use pact-jvm because of all of these dependencies. scala, kotlin, groovy, aws sdk, gson, kotson, jackson, jsonpath, netty, etc.

I had to exclude some libs in `build.gradle` dependencies of pact-jvm because they clashed with project ones.

* Do you plan to eradicate kotlin, scala and groovy from dependencies?
* Did you thought about distributing pact-jvm as shadow jars (aka fat jars)

------------------------
* [list of dependencies for consumer test](https://gist.github.com/tmszdmsk/edb99e342c9133babbbb4fae2db5d749#file-pact-jvm-consumer-dependencies-txt)
* [list of dependencies for provider test](https://gist.github.com/tmszdmsk/edb99e342c9133babbbb4fae2db5d749#file-pact-jvm-provider-dependencies-txt)

uglyog:
There is no plan to eradicate Kotlin, the plan is to replace Scala with Kotlin.

tmszdmsk:
Do you need help on that? What module can I start with?

uglyog:
@tmszdmsk pact-jvm-matchers is a core component that is used by everything else, so converting that would be the place to start.

The other one is converting the Groovy code in pact-jvm-model.

PR #634, opened 2018-01-30
groovy2kotlin pact-jvm-model: MatchingRulesImpl

tmszdmsk:
Related to #439.

Since there is less Groovy than Scala code that we want to replace with Kotlin i picked `MatchingRulesImpl.groovy` as a good candidate for rewrite to Kotlin.

Notes:
* I had to change two tests that depended on specific format used by Groovy `.toString()` methods to serialize object, is anyone depending on this specific groovish format?
* I see a lot of things that are wrong with this class but this PR is about "blindly" rewriting to kotlin
* `PactSerialiserSpec` had to be changed because somehow equality comparison changed and `TypeMatcher` from Kotlin didn't equal `new TypeMatcher()` from Groovy (which is privileged to call private constructors ;)  

Disclaimer:
* I am not a Kotlin master so probably not all things are idiomatic ;)

What do you think?


pact-jvm-consumer-groovy/src/test/groovy/au/com/dius/pact/consumer/groovy/PactResultSpec.groovy
```diff
@@ -134,7 +134,7 @@ class PactResultSpec extends Specification {
           '\tpath: /path\n' +
           '\tquery: [:]\n' +
           '\theaders: [Content-Type:application/json]\n' +
-          '\tmatchers: MatchingRules(rules=[body:Category(name=body, matchingRules={}), path:Category(name=path, matchingRules={})])\n' +
+          '\tmatchers: MatchingRules(rules={body=Category(name=body, matchingRules={}), path=Category(name=path, matchingRules={})})\n' +
```
tmszdmsk: `.toString` formatting is different in kotlin.

pact-jvm-model/src/test/groovy/au/com/dius/pact/model/PactSerialiserSpec.groovy
```diff
@@ -39,12 +39,12 @@ class PactSerialiserSpec extends Specification {
     provider = new Provider('test_provider')
     consumer = new Consumer('test_consumer')
     def requestMatchers = new MatchingRulesImpl()
-    requestMatchers.addCategory('body').addRule('$.test', new TypeMatcher())
+    requestMatchers.addCategory('body').addRule('$.test', TypeMatcher.INSTANCE)
```
tmszdmsk: calling private constructor of TypeMatcher was breaking equality test. Strange thing is that it worked before ;)

pact-jvm-consumer-groovy/src/test/groovy/au/com/dius/pact/consumer/groovy/PactResultSpec.groovy
uglyog: This does raise the issue of the output changing, which is mostly used in logging. Hopefully nobody depends on the format of the matchers which are output to the log files.

pact-jvm-model/src/test/groovy/au/com/dius/pact/model/PactSerialiserSpec.groovy
uglyog: oh, yeah, that is wrong. TypeMatcher is a singleton so the constructor should have never been called, but you can call private things in groovy.

uglyog:
All good, thank you. 

There are some improvements that can be made now that it is Kotlin source (like using a data class, which provides the copy, equals and hashcode methods by default), and cleanup some of the nested if statements. I'll do that as soon as I get some spare time.

PR #637, opened 2018-02-02
Add support for multipart/mixed content type

cpul:
- Refactored name of Multipart matcher to reflect that it supports more than 1 type of multipart request
- Added a matcher for 'multipart/mixed' request body
- Updated tests to support the other side of the GMT


uglyog:
Thanks for the PR.

I'm assuming `multipart/mixed` is a more general form of multipart bodies than a form post?

cpul:
No problem. The `multipart/mixed` content type is useful for sending something that contains multiple inline content-types. 

I was actually thinking that I could have taken an even more general approach to supporting other `multipart/*` content types, but I didn't have enough time to investigate how it would work with the current Multipart Matcher class. Perhaps a future PR!

Cheers.

PR #648, opened 2018-03-07
Dependency Cleanup

slu-it:
When I first started using Pact I noticed that the dependency footprint of the core modules (matchers and model) was unusually big. Having all modern JVM language libraries already on the classpath, it seamed odd to also have things like Apache Commons Lang and Guava.

I started to experiment which of these dependencies are actually needed and it turns out with a little moving around and removing some dependencies the footprint can be reduced quite a bit.

I split this change into 3 commits:

1. Removing unused dependencies from the `pact-jvm-model` module and moving some into the modules actually using them.
2. Moving the `Mock*ProviderConfig` classes from the central `pact-jvm-model` into the consumer, since they are no model classes and only consumer code actually needs them.
3. Removed the dependency on the `scala-compiler` for each Scala using module. The compiler should not generally be a dependency of any library unless that library actually needs to compile Scala code on the fly (e.g. REPL). Instead of the whole compiler and all of its transitive dependencies, only the `scale-xml` module, which is actually needed, should be used.

I still have one headache: Why is the `pact-jvm-model` depending on the `pact-broker` client library? Having that dependency in the most central module of the whole project doesn't feel right to me. It is correct, that both consumers and provider _might_ be using the broker and therefore need to communicate with it. But that should always be an opt-in dependency on a higher module level. If you like the changes I provided here, I might be looking into an elegant solution for that as well.


uglyog:
Thanks for the PR, I'm going to review all the changes now.

Most of the issues are due to result of refactoring and moving the code around over the years, so if things look like they are unnecessary (like the model project depending on the pact broker one), it is not by design so we can just change that.

slu-it:
Ok, then I'll look into that next

Issue #632, opened 2018-01-24
Spring boot and @PactBroker is not working as expected with system properties


steam0:
In Spring Boot you can provide an application.properties/application.yml file that contains system properties and environment properties.

I use some of the system variables in my test

**PactVerificationTest.java**
```
@RunWith(SpringRestPactRunner.class)
@Provider("someprovider")
@IgnoreNoPactsToVerify
@PactBroker(
        protocol = "https",
        host = "${pactbroker.host}", port = "${pactbroker.port}",
        authentication =  @PactBrokerAuth(username = "${pactbroker.auth.username}", password = "${pactbroker.auth.password}")
)
@SpringBootTest(webEnvironment = SpringBootTest.WebEnvironment.DEFINED_PORT)
public class PactVerificationTest {

    @TestTarget
    public final Target target = new HttpTarget(50519);

    @MockBean
    private EmailController emailController;

    @Before
    public void beforeTests() {
        doNothing().when(emailController).sendEmail(any(Email.class));
    }

    @State({"default", "extra"})
    public void toDefaultState() {
        System.out.println("Now service in default state");
    }
}
```

And then I name these variables in the configuration file

**application.yml**
```
pactbroker:
  host: "my.host.name.org"
  port: "443"
  auth:
    username: "myusername"
    password: "mypassword"
```

**Unfortunately this does not work and I get a RuntimeError**
`java.lang.RuntimeException: Could not resolve property "pactbroker.host" in the system properties or environment variables and no default value is supplied`

**So it seems as it wont read the variables.**

**Versions:**
Spring Boot 2.0.0.M7
au.com.dius:pact-jvm-provider-spring_2.12 3.5.11

uglyog:
The Pact runners look for the property values in the Java system properties, while the properties you supply to Spring are defined in the Spring context. That is why they are not found.

I don't recall a way to make the Spring context properties available as a Java system property.

uglyog:
This has now been implemented. I have a test project https://github.com/uglyog/pact-springboot-test with this working nicely.

steam0:
Awesome! Thank you so much, @uglyog !

uglyog:
3.5.14 has now been released

PR #649, opened 2018-03-08
Made the PactBroker host and port overridable using system properties.

prashant-ee:
I believe resolution order for the PactBroker annotation attributes (host, port and others) should be reversed. Ideally I would like to specify value for host and port in code but still have ability to override the values through system properties depending on the environment it is executed. 

This will allow different environment (local and Jenkins/CI) to have different values for host and port without changing anything in the code. 

The problem with earlier (existing) model is if say Jenkins/CI is setup to use particular pact broker values, that setup can very easily be overridden by committing a code value for these attributes. This reversal is particularly useful for tags to be fetched from Pact Broker (tag changes not implemented yet) allowing developer to run test locally with latest tag and Jenkins can run with all environment tags (UAT/PROD etc) .  Tag change might also go unnoticed as build will not go red. 

I can made similar changes to other attributes if you think this is a useful change.  

Also let me know whether any documentation is required to be updated and guide me as per how to do that.

### DIFF
--- a/pact-jvm-provider-junit/src/main/java/au/com/dius/pact/provider/junit/loader/PactBroker.java
+++ b/pact-jvm-provider-junit/src/main/java/au/com/dius/pact/provider/junit/loader/PactBroker.java
@@ -23,12 +23,12 @@ public @interface PactBroker {
     /**
      * @return host of pact broker
      */
-    String host() default "${pactbroker.host:}";
+    String host() default "";
 
     /**
      * @return port of pact broker
      */
-    String port() default "${pactbroker.port:}";
+    String port() default "";
 
     /**
      * HTTP protocol, defaults to http
--- a/pact-jvm-provider-junit/src/main/java/au/com/dius/pact/provider/junit/loader/PactBrokerLoader.java
+++ b/pact-jvm-provider-junit/src/main/java/au/com/dius/pact/provider/junit/loader/PactBrokerLoader.java
@@ -59,7 +59,7 @@ public class PactBrokerLoader implements PactLoader {
   }
 
   public PactBrokerLoader(final PactBroker pactBroker) {
-      this(pactBroker.host(), pactBroker.port(), pactBroker.protocol(), Arrays.asList(pactBroker.tags()));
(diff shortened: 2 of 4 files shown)


uglyog:
I disagree with this change. I understand the convenience it provides for your use case, but it means that for other users who specify the value in code, they could have it replaced at runtime with a system property, which would be mostly unexpected. I can already foresee the issues that will be raised for this.

The expressions used with the annotations are resolved with an expression parser that was designed to be pluggable. I think a better solution would be for you to be able to override the default expression resolver with one that changes the order.

prashant-ee:
Thanks for the reply. I submitted the pull request as I think this will be useful for all. But I take sure point the new behaviour can come as a surprise for existing users of the system.  

I was not aware of the expression parser, Can you point me towards it in the code?

Thanks.

uglyog:
The default implementation is here: https://github.com/DiUS/pact-jvm/tree/master/pact-jvm-provider-junit/src/main/java/au/com/dius/pact/provider/junit/sysprops

We just need to implement a way for you to override it. I also need to change this for #632

uglyog:
You can currently override it on the pact broker annotation: https://github.com/DiUS/pact-jvm/blob/master/pact-jvm-provider-junit/src/main/java/au/com/dius/pact/provider/junit/loader/PactBroker.java#L59

prashant-ee:
Thanks for sharing. I think I will have to explore a bit more, currently I am not sure how can this allow people to have there local machine (dev machine) setup regarding pact broker and still not affect things on build server (control that through system properties, regardless of settings in the code). But I guess this can help. Thanks.

uglyog:
You can create your own property resolver that does exactly what you want.

TimothyJones:
Can you share why you want to have separate brokers for development and production? For the common use cases, I think pact tags (with one broker) would solve your problem.

rholshausen:
The default expressions have a fallback value, and this will do what you want.

For example:

```java
@PactBroker(host="${pactbroker.hostname:dev-boker.local}")
```

In this case, it will use the system property value of `pactbroker.hostname`, and if it is not defined, it will use `dev-boker.local`.

Issue #662, opened 2018-04-08
Getting a "No such property: providerVersion" when executing mvn pact:verify on JDK6 project


jaspreetParhar:
Hello iam trying to use the pact-jvm-provider-maven_2.10 plugin version: 2.4.19 for a jdk 6 project.
When iam trying to run mvn pact:verify iam getting following error:

**Failed to execute goal au.com.dius:pact-jvm-provider-maven_2.10:2.4.19:verify (default-cli) on project mbconnect.sbb.cpd.ws: Execution default-cli of goal au.com.dius:pact-jvm-provider-maven_2.10:2.4.19:verify failed: No such property: providerVersion for class: au.com.dius.pact.provider.ProviderVerifier** 

I tried to find a similar failure on inet but there was nothing. Please can you help me.

Thanks in advance.

uglyog:
Ok, looks like a commit was missed when back-porting the fixes from the 3.5.x version.

I'll cherry-pick the associated commits for you, and release a new version.

jaspreetParhar:
Thank you very much for the fix.

uglyog:
First have to get the build passing, and then do a release. ;-)

I am unable to test on JDK 6, so you will need to verify it.

jaspreetParhar:
Iam on it ;) will let you know in a while.

jaspreetParhar:
Hello soo I tried to build the plugin but getting following error!!!

![image](https://user-images.githubusercontent.com/15687935/38538025-3b002d7a-3c92-11e8-86f6-f10567b15288.png)

uglyog:
Can you have a look at the test report to see why that test is failing?

jaspreetParhar:
Hello following test PactBodyBuilderSpec is failing. 
This is the caused by message which I am seeing. 
au.com.dius.pact.consumer.groovy.PactBodyBuilderSpec.Guard Against Field Names That Don't Conform To Gatling Fields(PactBodyBuilderSpec.groovy:414)

jaspreetParhar:
I have noticed an other problem as well au.com.dius.pact.model.PactFragment does not exists and this cause errors too.

uglyog:
I'm not getting any errors about missing PactFragment. I have got the build passing, so I can do a release.

jaspreetParhar:
Hello soo I am able to get the build passing if I am changing the version from 2.4.20 --> 2.4.19 in the build.gradle file direct under the root pact-jvm project. 
[...]

PR #663, opened 2018-04-13
Fix to build pact-jvm-maven plugin #662

jaspreetParhar:
I opened an issue https://github.com/DiUS/pact-jvm/issues/662. There were several changes and I had to test if the changes do work. 
I could not test because the pact-jvm-maven-plugin could not be build due to this error:
image

After this fix the gradle build is successful.


uglyog:
Thanks for the PR.

PR #666, opened 2018-04-17
Add in the capability to register a pact file to be accessible from the main server API.

groboclown:
This change allows the primary pact-jvm server to also run as the entry point to the pacts.  The original use of the server will still work.

With this enhancement, it's now easier to use the pact-jvm server to mock other services.  If you have service A that uses services B and C, then you can run integration or pact tests against service A without needing to run service B and C.  Instead you can point service A to the pact-jvm server.  Without this change, the setup for service A requires dynamic and potentially difficult changes to correctly point its service clients to the correct pact-jvm server ports.  When the pact-jvm server is run as a docker container along with service A, then the setup is even easier by aliasing the pact-jvm server.

The new enhancement only works if the mocked services have distinct REST path prefixes, such as `/functionA/create` and `/functionB/create`.  The RequestRouter will direct the REST request to the pact service with the registered path prefix.

### DIFF
--- a/pact-jvm-server/README.md
+++ b/pact-jvm-server/README.md
@@ -98,16 +98,21 @@ The following actions are expected to occur
 
 ### /create
 
-The client will need `POST` to `/create` the generated `JSON` interactions, also providing a state as a query parameter.
+The client will need `POST` to `/create` the generated `JSON` interactions, also providing a state as a query parameter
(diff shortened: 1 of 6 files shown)


uglyog:
Thanks for the PR. That looks useful.

PR #671, opened 2018-04-26
Added methods newJsonArrayMinLike and newJsonArrayMaxLike to LambdaDsl

softarn:
Currently there is no option to create an array on root level with a minimum or maximum number of objects using the LambdaDsl class. This commit adds this functionality by setting number of examples and adding a rule to the pactDslJsonArray object.


yurido:
men bra jobbat då!

PR #673, opened 2018-05-02
Don't run servers on commonly used ports during tests

pkubowicz:
Tests failed if I had a Tomcat instance running on 8080


uglyog:
Good point, thanks for the PR

PR #682, opened 2018-05-20
evaluate @State annotations from implemented interfaces (fixes #646)

### LINKED ISSUE #646: Process `@State` annotation from implemented interfaces
I have a feature request that would help us to write modular pact provider tests while still making sure that all interactions of a pact are being validated.

**Background:**
Let's say we have a UserService implemented with Spring Boot and a UI implemented with Angular. The consumer tests in the UI build are creating a pact file defining all interactions between UserService and UI.

On the provider side we have multiple RestControllers, each providing endpoints for a subset of the interactions defined in the pact. 

The default way to create a provider test with pact-jvm is to create a single provider test class for the whole pact file. However, this test class may become quite large and is a pain to work on in parallel. 

**Current Workaround:**
We would like to separate concerns and create a separate provider test class per RestController, each testing a subset of the interactions. Currently, we are doing that by using `@PactFilter`. This apparently does not work yet in combination with `@SpringBootTest` (see #572). Also, using `@PactFilter` has the disadvantage that we cannot be sure that all interactions defined in the pact are actually being tested.

**Feature Idea:**
Process `@State` annotations not only on methods of the provider test class itself, but also on (default) methods of all implemented interfaces. 

Then we could do something like this:

```
public interface Controller1ProviderTest {
  
  UserRepository userRepositoryMock();

  @State("user 42 exists")
  default void shouldGetUser() {
    when(userRepositoryMock().findOne(42))
        .thenReturn(new User("Zaphod"));
  }

  // more @State methods for all states that Controller1 needs
  
}
```

```
@Provider("userservice")
@PactBroker(...)
@SpringBootTest(...)
@RunWith(SpringRestPactRunner.class)
public class UserServiceProviderTest implements Controller1ProviderTest, Controller2ProviderTest, ... {
  
  @TestTarget
  public final HttpTarget target = new HttpTarget(...);

  @MockBean
  private UserRepository userRepository;

  @Override
  UserRepository userRepositoryMock(){
    return userRepository;
  }

  // implement mock factory methods of all implemented interfaces
  
}
```

`UserServiceProviderTest` then acts as the single entry point for testing all interactions of the pact while we can happily create a separate test class (in the form of an interface with default methods) for each controller.

Does that make any sense? :)

thombergs:
In this PR `@State` annotations from default methods in implemented interfaces will be evaluated as well.

This makes the use case possible that is described in #646, thus allowing to modularize the provider tests for pacts with more than just a few interactions.


thombergs:
When implementing this, I noticed that multiple `@State` annotations with the same state don't create an exception. I would have expected an exception saying that the same `@State` is not allowed more than once. Probably either the first or the last method is then executed? Or both? 

Shouldn't there be a validation that we have each state only once?

thombergs:
btw, if this is accepted, I'll gladly try to do the same for the junit5 module

uglyog:
Just checked the code with respect to multiple `@State` annotations, and I think that may have been by design. It will run all methods annotated with the correct state annotation.

thombergs:
I'll have a look at the junit5 module some time to implement the same functionality there, too.

PR #685, opened 2018-05-23
Added support for inherited `@State` annotations

thombergs:
To enable the provider test style described in #646 with JUnit 5, I moved the `PactVerificationStateChangeExtension` from `beforeEach` to `beforeTestExecution`.

This makes sure that the state changes are called **after** all `@BeforeEach` methods have been called. This, in turn, makes it possible to use the `@BeforeEach` methods to populate class members which can then be used in the `@State`-annotated methods.

Without this change, when you set a class member within a `@BeforeEach` method and then tried to access it in a `@State` method, you would get a `NullPointerException`.

This PR also adds a test that asserts this behavior works.


rholshausen:
Good change!

PR #688, opened 2018-05-28
made default content-type more lenient to common variants

jan-berge-ommedal:
PactDslResponse.body() now defaults to match content-type against `application/json;\s?charset=(utf|UTF)-8` instead of requiring the exact value `application/json; charset=UTF-8`.

This makes the default work with producers running for example jersey and jetty which uses the compact format `application/json;charset=utf-8`.

### DIFF
--- a/pact-jvm-consumer/src/main/java/au/com/dius/pact/consumer/dsl/PactDslResponse.java
+++ b/pact-jvm-consumer/src/main/java/au/com/dius/pact/consumer/dsl/PactDslResponse.java
@@ -2,9 +2,7 @@ package au.com.dius.pact.consumer.dsl;
 
 import au.com.dius.pact.consumer.ConsumerPactBuilder;
 import au.com.dius.pact.model.OptionalBody;
-import au.com.dius.pact.model.Pact;
 import au.com.dius.pact.model.PactFragment;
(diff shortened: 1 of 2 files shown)


abubics:
This is a good step.

Ideally, Pact shouldn't be making assertions about anything that's not explicit in the test, though :<

pact-jvm-consumer/src/main/java/au/com/dius/pact/consumer/dsl/PactDslResponse.java
```diff
@@ -21,18 +19,18 @@
...
 import javax.xml.transform.TransformerException;
-import java.util.Arrays;
-import java.util.Collections;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.function.Supplier;
 
 public class PactDslResponse {
     private static final String CONTENT_TYPE = "Content-Type";
+    static final String DEFAULT_JSON_CONTENT_TYPE_REGEX = "application/json;\\s?charset=(utf|UTF)-8";
```
rholshausen: I think the regex should accept all charsets, not just UTF-8. On Windows it could be one of the Windows specific charsets.

PR #717, opened 2018-06-23
java8 dsl only needs the java model dependency

signed:
I want to use the java8 dsl for my junit5 tests without pulling in junit 4.12 dependencies.

### DIFF
--- a/pact-jvm-consumer-java8/build.gradle
+++ b/pact-jvm-consumer-java8/build.gradle
@@ -7,5 +7,5 @@ repositories {
 }
 
 dependencies {
-  compile project(":pact-jvm-consumer-junit_${project.scalaVersion}")
+  compile project(":pact-jvm-consumer_${project.scalaVersion}")
 }


signed:
Just realizing that the junit 5 modules depend on `compile project(":pact-jvm-consumer-junit_${project.scalaVersion}")` that brings in junit 4.12 :(.

Are you open to breaking this dependency by either creating a new module `*-junit4`or `*-junit-base`?

uglyog:
Definitely need to correct these dependencies. It is mainly re-using the annotations from the JUnit4 module, so I think we can move those to the consumer module, as they don't have any JUnit4 specific classes.

signed:
At first glance there is at leas `JUnitTestSupport` in `pact-jvm-consumer-junit` that is used in `pact-jvm-consumer-junit5`.

Duplicate the code?
Introduce a `pact-jvm-consumer-junit-base` module and put it there?

uglyog:
I wouldn't create a new module just for a few classes, we can put them in `pact-jvm-consumer`

signed:
I played with this a bit over the last days.
Here are some findings.

Moving the  `Pact` annotation into pact-jvm-consumer results in a clash with a class in `ConsumerPactRunner` which imports Pact from model.
Moving `JUnitTestSupport` into pact-jvm-consumer kotlin src results in PactDslWithProvider not being resolvable because src/java is excluded from the import in the gradle build.
Removing the exclude results in a cyclic dependency in the gradle configuration between the different source folders.
This could probably be resolved by reducing the number of involved languages, and porting the other code into that language.
Are there technical requirements for that many languages?
Do you have a preferred language for this project you want to move to?

I tried the approach with an additional module, but ended up with a kotlin class that can not be resolved in the build.
Looks to me like a gradle configuration issue, but I do not understand the setup enough.
I pushed the spike to a branch in my repository https://github.com/signed/pact-jvm/tree/additional_module. 
Even with only a few classes in the new module, this feels the more promising approach to me.

uglyog:
I've moved the classes and removed the dependency in 7c996e71, so we can close this PR

signed:
thx for cleaning up the dependencies

PR #719, opened 2018-06-25
Use kotlin-stdlib-jdk8 insted of kotlin-stdlib-jre8.

yokotaso:
Because kotlin-stdlib-jre8 is depcated. It seems good to use  kotlin-stdlib-jdk8.


uglyog:
Thanks for the PR!

PR #731, opened 2018-07-11
Add system properties configuration to plugin

### LINKED ISSUE #730: pact-jvm-provider-maven does not allow setting pact.verifier.publishResults to true
As `mvn pact:verify` is its own goal, it does not seem that it sees `pact.verifier.publishResults` is set to true from the surefire-plugin. So, `mvn pact:verify` yields
```
[WARNING] Skipping publishing of verification results as it has been disabled (pact.verifier.publishResults is not 'true')
```
Fix would need to allow such configuration in pom.xml.

SnewsButton:
Fixes #730.
* Allows systemPropertyVariables tag within pact-jvm-provider-maven_2.12 configuration.
* Sets the system property when executed
* updated  project root's` Readme.md` as pact-maven plugin can publish pacts and referenced project has not been updated in a while and has some problems with requesting pact files.

### DIFF
--- a/README.md
+++ b/README.md
@@ -104,8 +104,6 @@ You want to look at: [pact gradle plugin](pact-jvm-provider-gradle)
 
 You want to look at: [pact maven plugin](pact-jvm-provider-maven)
 
-For publishing pacts to a pact broker, have a look at https://github.com/warmuuh/pactbroker-maven-plugin
-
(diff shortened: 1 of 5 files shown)


uglyog:
Thanks for the PR.

Just a note on Spock tests with system properties, Spock has a `@RestoreSystemProperties` annotation that can reset the system properties so one test won't pollute the properties for others. I'll add that on to your test.

PR #750, opened 2018-08-04
Add teardown support to JUnit provider

tinexw:
Hi, 

this is a first attempt at adding teardown support to the Junit provider. Are you ok with this approach in general or would you prefer a different one?
Where should I add the tests? There are groovy, java and kotlin folders and I don't know where to put them.

### DIFF
--- a/pact-jvm-provider-junit/src/main/java/au/com/dius/pact/provider/junit/InteractionRunner.java
+++ b/pact-jvm-provider-junit/src/main/java/au/com/dius/pact/provider/junit/InteractionRunner.java
@@ -236,9 +236,10 @@ public class InteractionRunner extends Runner {
             target.testInteraction(pact.getConsumer().getName(), interaction, source);
           }
       };
-      statement = withStateChanges(interaction, testInstance, statement);
+      statement = withStateChanges(StateChangeAction.SETUP, interaction, testInstance, statement);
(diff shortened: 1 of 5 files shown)


uglyog:
This is a good start. You can write the tests in the language you are most comfortable with.

tinexw:
I added a test. Let me know if anything else is missing.

tinexw:
@uglyog Did you have a chance to look at my changes?

uglyog:
The only thing that is missing is to update RunStateChanges so it invokes the teardown method afterwards.

tinexw:
It's invoked in the InteractionRunner. Does it really need to be invoked somewhere else?

uglyog:
It is invoked, but only runs before the interactions. It needs to also run the teardowns after. But I will sort that out for you.

uglyog:
It was actually running the teardown first, then the setup, then the interaction. I have refactored it a bit, and implemented the execution order correctly. The missing bit was https://github.com/DiUS/pact-jvm/blob/v3.5.x/pact-jvm-provider-junit/src/main/java/au/com/dius/pact/provider/junit/RunStateChanges.java#L27

uglyog:
3.5.22 has been released with this feature

tinexw:
Thanks!

PR #764, opened 2018-08-26
Pact broker filter by consumers

carlzogheib:
Hey all,

This is to add the ability to filter provider pact test runs that are loaded via the `PactBrokerLoader` by specific consumers (comma-separated list of consumers)

This, as is done for tags, can be read through a system property `pactbroker.consumers` in cases where we're only interested in testing selected consumer pacts for a provider.

Use case: a consumer has updated their pact and wants to validate it with the provider - a provider pact test run is then triggered for only this pact, ignoring the provider's other pacts with other consumers.

Cheers

### DIFF
--- a/pact-jvm-provider-junit/src/main/java/au/com/dius/pact/provider/junit/loader/PactBroker.java
+++ b/pact-jvm-provider-junit/src/main/java/au/com/dius/pact/provider/junit/loader/PactBroker.java
@@ -41,6 +41,12 @@ public @interface PactBroker {
      */
     String[] tags() default "${pactbroker.tags:latest}";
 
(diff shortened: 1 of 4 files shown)


carlzogheib:
Also took the opportunity to fix some tests for `pactbroker.tags` not testing the right things

uglyog:
Awesome! Thanks for the PR, this is a great addition.

carlzogheib:
Hey @uglyog any news on when this can be released?

uglyog:
I'll sort out a release this weekend.

uglyog:
3.5.22 has been released

PR #776, opened 2018-09-23
Stop pact-jvm-consumer-java8 being dependent on JUnit 4

pkubowicz:
Unnecessary dependency caused problems when using JUnit 5.
Actually the code is completely independent of JUnit.

Additionally remove repetitions from root build.gradle


uglyog:
Thanks for the PR.

uglyog:
Yes, we can merge to 3.5.x. Probably is a good idea.

PR #777, opened 2018-09-23
Stop pushing compile dependencies to all projects

pkubowicz:
Use only transitive dependencies for compile dependencies.
Pushing dependencies to subprojects introduces a risk of polluting published POM with unneeded dependencies.
    
Keep pushing test dependencies as it does not affect library clients.


uglyog:
Thanks for the PR

PR #787, opened 2018-10-07
Fix grammar of heading for Clojure usage instructions

devurandom:


uglyog:
Clojure (being a Lisp dialect), uses infix notion as well as brackets, so the correct form of `I use Clojure` in Clojure is `(use I Clojure)`. It was a joke :-D

PR #790, opened 2018-10-10
Add @DelegatesTo on the groovy dsl pact builders to allow IDE code co…

tvrmsmith:
…mpletion.

This may not be the most ideal solution to this problem but it was the most straightforward.


uglyog:
Nice! I didn't know about the `@DelegatesTo` annotation.

pact-jvm-consumer-groovy/src/main/groovy/au/com/dius/pact/consumer/groovy/BaseBuilder.groovy
```diff
@@ -4,16 +4,20 @@ package au.com.dius.pact.consumer.groovy
...
-    def build(Closure closure) {
-        closure.delegate = this
-        closure.resolveStrategy = Closure.DELEGATE_FIRST
-        closure.call()
-    }
+  def build(Closure closure) {
+    callWithThisAsDelegate(closure)
+  }
+
+  def callWithThisAsDelegate(Closure closure) {
```
uglyog: I don't understand the need for this method. `call` invokes `build` which in turn invokes `callWithThisAsDelegate`.

PR #796, opened 2018-10-17
Send detailed verification results to pact broker

tinexw:
This is a first attempt to implement the following request mentioned in the pact-jvm slack channel:

> Hi, The Matrix on Pact Broker shows the high-level verification results, is there any way in Pact Broker where a user can see detailed verification result. 

(See https://pact-foundation.slack.com/archives/C9VPNUJR2/p1539079990000100)

I only tested it manually for Maven and Junit 4 and I didn't add any automated tests yet because I would like to get some feedback on the approach first. The formatting is far from perfect but maybe it' sufficient, what do you think?

Maven:
```
{
  "providerName": "user-service",
  "providerApplicationVersion": "1.0.0-SNAPSHOT",
  "success": false,
  "verificationDate": "2018-10-17T22:12:22+00:00",
  "testResults": "{Verifying a pact between messaging-app and user-service - A request for a non-existing user  Given default returns a response which has status code 404=expected status of 404 but was 200}”
}
```

Junit 4:
```
{
  "providerName": "user-service",
  "providerApplicationVersion": "0.0.0",
  "success": false,
  "verificationDate": "2018-10-17T22:49:24+00:00",
  "testResults": "[messaging-app - A request for a non-existing user(de.kreuzwerker.cdc.userservice.GenericStateWithParameterContractTest): \n0 - expected status of 404 but was 200]"
}
```

### DIFF
--- a/pact-jvm-pact-broker/src/main/kotlin/au/com/dius/pact/pactbroker/PactBrokerClient.kt
+++ b/pact-jvm-pact-broker/src/main/kotlin/au/com/dius/pact/pactbroker/PactBrokerClient.kt
@@ -27,12 +27,16 @@ abstract class PactBrokerClientBase(val pactBrokerUrl: String, val options: Map<
     docAttributes: Map<String, Map<String, Any>>,
     result: Boolean,
     version: String,
+    testResults: String? = null,
     buildUrl: String? = null
   ): Result<Boolean, Exception> {
     val halClient = newHalClient()
     val publishLink = docAttributes.mapKeys { it.key.toLowerCase() } ["pb:publish-verification-results"] // ktlint-disable curly-spacing
     return if (publishLink != null) {
       val jsonObject = jsonObject("success" to result, "providerApplicationVersion" to version)
+      if (testResults != null) {
+        jsonObject.add("testResults", testResults.toJson())
+      }
       if (buildUrl != null) {
         jsonObject.add("buildUrl", buildUrl.toJson())
       }
--- a/pact-jvm-provider-junit/src/main/kotlin/au/com/dius/pact/provider/junit/InteractionRunner.kt
+++ b/pact-jvm-provider-junit/src/main/kotlin/au/com/dius/pact/provider/junit/InteractionRunner.kt
@@ -136,13 +136,16 @@ open class InteractionRunner<I>(
   // Running
   override fun run(notifier: RunNotifier) {
     var allPassed = true
+    val failures = mutableListOf<String>()
     for (interaction in pact.interactions) {
       val description = describeChild(interaction)
       notifier.fireTestStarted(description)
       try {
         interactionBlock(interaction, pactSource, testContext).evaluate()
       } catch (e: Throwable) {
-        notifier.fireTestFailure(Failure(description, e))
+        val failure = Failure(description, e)
+        failures.add(failure.toString())
+        notifier.fireTestFailure(failure)
         allPassed = false
       } finally {
         notifier.fireTestFinished(description)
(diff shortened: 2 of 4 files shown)


uglyog:
I have a question about the purpose of the `testResults` attribute. It is just a string, but it looks like it is intended to be a collection of the individual test results that contributed to the overall result. In that case, I would like the format to be changed to something more appropriate. This is probably a @bethesque question.

The implementation is also is only sending the failures, so if that is the case then we should probably rename it to `failedTestResults` or include all the results. 

@bethesque thoughts?

bethesque:
Currently, the ruby impl is pushing a hash to testResults. I've deliberately left it without a defined format, because I wasn't sure what the usecase was yet. It's expecting a JSON object @tinexw. I'd recommend a hash, as that's more extensible. If you provide a string or an array, you can't add extra information later. I'll see if I can find an example from the ruby code.

bethesque:
Here's an example of what the ruby impl generates.

```json
{
  "success": false,
  "providerApplicationVersion": "1.2.3",
  "testResults": {
    "tests": [
      {
        "testDescription": "has status code 200",
        "testFullDescription": "Verifying a pact between Foo and Bar A retrieve thing request with GET /thing returns a response which has status code 200",
        "status": "passed",
        "interactionProviderState": null,
        "interactionDescription": "a retrieve thing request",
        "actualStatus": 200
      },
      {
        "testDescription": "has a matching body",
        "testFullDescription": "Verifying a pact between Foo and Bar A retrieve thing request with GET /thing returns a response which has a matching body",
        "status": "failed",
        "interactionProviderState": null,
        "interactionDescription": "a retrieve thing request",
        "exception": {
          "class": "RSpec::Expectations::ExpectationNotMetError",
          "message": "Actual: {\"name\":\"Thing 2\"}\n\nDiff\n--------------------------------------\nKey: - is expected \n     + is actual \nMatching keys and values are not shown\n\n {\n-  \"name\": \"Thing 1\"\n+  \"name\": \"Thing 2\"\n }\n\nDescription of differences\n--------------------------------------\n* Expected \"Thing 1\" but got \"Thing 2\" at $.name"
        },
        "actualBody": {
          "name": "Thing 2"
        },
        "differences": [
          {
            "description": "Expected \"Thing 1\" but got \"Thing 2\" at $.name"
          }
        ]
      },
      {
        "testDescription": "\"Content-Type\" which equals \"application/json\"",
        "testFullDescription": "Verifying a pact between Foo and Bar A retrieve thing request with GET /thing returns a response which includes headers \"Content-Type\" which equals \"application/json\"",
        "status": "failed",
        "interactionProviderState": null,
        "interactionDescription": "a retrieve thing request",
        "exception": {
          "class": "RSpec::Expectations::ExpectationNotMetError",
          "message": "Expected header \"Content-Type\" to equal \"application/json\", but was \"application/hal+json\""
        },
        "actualHeaders": {
          "Content-Type": "application/hal+json",
          "Content-Length": "18"
        }
      }
    ],
    "summary": {
      "testCount": 3,
      "failureCount": 2
    },
    "metadata": {
      "warning": "These test results use a beta format. Do not rely on it, as it will definitely change.",
      "pactVerificationResultsSpecification": {
        "version": "1.0.0-beta.1"
      }
    }
  }
}
```

I'm not tied to this format, but I would want some kind of summary, and an array of results, where each result could be tied back to it's interaction by description and provider state, so that we can map a pass/fail back to each interaction in a given pact, and display it in the UI.

This example is for a v2 pact - a v3 pact would obviously have a more complex provider state structure.

tinexw:
Thanks! I will look into it. I will close this PR for now because that will take me some time.

PR #797, opened 2018-10-21
Detect Gradle when setting default value of pact.rootDir

pkubowicz:


uglyog:
This is awesome! It's been bothering me for years :-D

PR #828, opened 2018-12-08
Feat: snapshot trimming on provider version

aplsup:
As a provider, I would like to be able to trim the -SNAPSHOT on my version during verification result publications


anto-ac:
Liking this!

PR #850, opened 2019-02-15
Correcting PactTestFor.port-doc, default is indeed 0

SchulteMarkus:
The documentation at PactTestFor.port says, port 8080 would be chosen as default. As can be seen in https://github.com/DiUS/pact-jvm/blob/master/pact-jvm-consumer-junit5/src/main/kotlin/au/com/dius/pact/consumer/junit5/PactConsumerTestExt.kt#L102 , indeed port 0 is chosen, which will cause the MockServer being started at a random, free port.


uglyog:
Good catch!

Issue #889, opened 2019-06-05
Successful verification is published incorrectly when using SpringRestPactRunner and @PactFilter


anto-ac:
**How to reproduce it**

1. Publish to the pact broker a pact between `consumer A` and `provider A` with more than one interaction (e.g. `@State("state1")` and `@State("state2")`)
2. Run the provider verification using the `SpringRestPactRunner` and filter only one interaction using `@PactFilter("state1")`

**Expected result**

The pact should not be verified successfully (ideally) but definitely a successful verification result should not be published

**Actual result**

The pact shows as verified successfully on the pact broker.

anto-ac:
If you could give me a hint on where the problem may be or what would need changing, I or one of my colleagues may give it a go at fixing this.

PR #890, opened 2019-06-06
fix: Don't send successfull verification result to the Broker if not all interactions are verified

aplsup:
…all interactions are verified


uglyog:
Thanks for the PR! Does this fix #889, or is there still more needed?

I will backport it into the 3.6.x branch.

aplsup:
Actually I was thinking if we need some sort of global aggregation of all the test class with a PactRunner. While this PR prevents false positives, in the case where the verification tests for a PACT are split among different classes (filtered) the verification won’t be considered successful even if it semantically is.

anto-ac:
@uglyog What do you think about @aplsup consideration above?

Also, when you reckon we could have a 3.6.8 release?

uglyog:
I have cherry-picked the changes onto the v3.6.x branch.

A global aggregation will work as long as the tests run in the same JVM. If parallel test execution is enabled, then it won't work.

uglyog:
I'll organise a release this weekend

Issue #912, opened 2019-07-21
NumberValue is serialized in the wrong way in 4.0.0-beta.4


gmariotti:
Hi team, I think I found a bug in `4.0.0-beta.4`. 
In `4.0.0-beta.3`, a json field `numberValue("version", 1)` was serialized in the pact as `"version": 1`, however, in `4.0.0-beta.4` the same field is serialized as `"version": { "value": "1"}`, making failing the provider test.

uglyog:
Is this with the Groovy DSL?

gmariotti:
no, it's with the `java8` dsl. I'm using the `LambdaDslObject.numberValue` method

uglyog:
4.0.0-beta.5 has been released

gmariotti:
Fixed, thank you so much

Issue #908, opened 2019-07-11
Pact Consumer 4.0.0-beta.3 generating numberValue as object instead of primitive 


xCoBaLTz:
When building the consumer pact file with 4.0.0-beta.3, using the LambdaDSL, I noticed that it converts NumberValue to a Object with a value attribute containing the number expected instead of just expecting the number itself.

Expected Value:
`{
    "number": 1
}`

Actual Value:
`{
   "number":  {
       "value":  1
   }
}`

The code block I am referring to is the following:
`LambdaDsl.newJsonBody(o -> o.numberValue("number", 1)).build()`

This belongs to the pact-jvm-consumer-java8 artifact. This was not the case in 4.0.0-beta.2, and was working as expected. 

If this is an expected behaviour going forward, would it be possible to add another method in the LambdaDSL package to support such a case.

uglyog:
I am unable to replicate this. The following code passes when run against the current master.

```java
        DslPart dslPart = LambdaDsl.newJsonBody(o -> o.numberValue("number", 1)).build();
        assertThat(dslPart.getBody().toString(), is("{\"number\":1}"));
```

igordezky-blackberyy:
With 4.0.0-beta, when the map returned from Interaction.toMap() is serialized the type is a Number. The default gson serialization of Number is { "value": <value> }. This can be addressed using registerTypeHierarchyAdapter(Number.class, ...) to provide custom serialization.

uglyog:
4.0.0-beta.5 has been released

xCoBaLTz:
Thank you 👍

PR #913, opened 2019-07-22
Fix gson number serialization #912, #908

igordezky-blackberyy:
With this I am able to generate the correct interactions format using junit5


uglyog:
Thanks for the PR. You'd think this is something GSon would do out of the box!

PR #915, opened 2019-07-23
feat: specify the pact tag if applicable

oswaldquek:
When provider tests are  run using the PactBroker, and multiple tags are
specified, it can be to figure out which verification was for which tag. This
is slightly annoying especially when verifications fail. This commit prints out
the tag if supplied.

Note: This is my first time doing anything in kotlin/groovy proper, so definitely do point out any bad practices.

### DIFF
--- a/provider/pact-jvm-provider-junit5/src/main/kotlin/au/com/dius/pact/provider/junit5/PactJUnit5VerificationProvider.kt
+++ b/provider/pact-jvm-provider-junit5/src/main/kotlin/au/com/dius/pact/provider/junit5/PactJUnit5VerificationProvider.kt
@@ -171,7 +171,7 @@ class PactVerificationExtension(
     val providerInfo = testContext.target.getProviderInfo(serviceName, pactSource)
     testContext.providerInfo = providerInfo
 
-    prepareVerifier(testContext, context)
+    prepareVerifier(testContext, context, pactSource)
     store.put("verifier", testContext.verifier)
 
     val requestAndClient = testContext.target.prepareRequest(interaction, testContext.executionContext ?: emptyMap())
@@ -185,7 +185,7 @@ class PactVerificationExtension(
     }
   }
 
-  private fun prepareVerifier(testContext: PactVerificationContext, extContext: ExtensionContext) {
+  private fun prepareVerifier(testContext: PactVerificationContext, extContext: ExtensionContext, pactSource: au.com.dius.pact.core.model.PactSource) {
     val consumer = ConsumerInfo(consumerName ?: pact.consumer.name)
 
     val verifier = ProviderVerifier()
@@ -194,7 +194,7 @@ class PactVerificationExtension(
     setupReporters(verifier, serviceName, interaction.description, extContext, testContext.valueResolver)
 
     verifier.initialiseReporters(testContext.providerInfo)
-    verifier.reportVerificationForConsumer(consumer, testContext.providerInfo)
+    verifier.reportVerificationForConsumer(consumer, testContext.providerInfo, pactSource)
 
     if (!interaction.providerStates.isEmpty()) {
       for ((name) in interaction.providerStates) {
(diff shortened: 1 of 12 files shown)


oswaldquek:
@uglyog could you have a look at this please?

provider/pact-jvm-provider-junit5/src/main/kotlin/au/com/dius/pact/provider/junit5/PactJUnit5VerificationProvider.kt
```diff
@@ -185,7 +185,7 @@ class PactVerificationExtension(
     }
   }
 
-  private fun prepareVerifier(testContext: PactVerificationContext, extContext: ExtensionContext) {
+  private fun prepareVerifier(testContext: PactVerificationContext, extContext: ExtensionContext, pactSource: au.com.dius.pact.core.model.PactSource) {
```
uglyog: Probably better to import the class

provider/pact-jvm-provider/src/main/groovy/au/com/dius/pact/provider/ProviderVerifier.groovy
```diff
@@ -1,13 +1,16 @@
...
 import au.com.dius.pact.core.model.FilteredPact
 import au.com.dius.pact.core.model.Interaction
 import au.com.dius.pact.core.model.Pact
 import au.com.dius.pact.core.model.PactReader
+import au.com.dius.pact.core.model.PactSource
 import au.com.dius.pact.core.model.UrlPactSource
 import au.com.dius.pact.core.pactbroker.PactBrokerClient
 import au.com.dius.pact.core.pactbroker.TestResult
 import groovy.util.logging.Slf4j
+import org.jetbrains.annotations.Nullable
```
uglyog: This is a Jetbrains annotation, but I don't think the library is included anywhere. Unless it comes along with Kotlin?

provider/pact-jvm-provider/src/main/kotlin/au/com/dius/pact/provider/reporters/AnsiConsoleReporter.kt
```diff
@@ -45,7 +45,7 @@ class AnsiConsoleReporter(
 
   override fun finaliseReport() { }
 
-  override fun reportVerificationForConsumer(consumer: IConsumerInfo, provider: IProviderInfo) {
+  override fun reportVerificationForConsumer(consumer: IConsumerInfo, provider: IProviderInfo, tag: String?) {
```
uglyog: Should use the tag here too, in case anyone is relying on the console output.

provider/pact-jvm-provider-junit5/src/main/kotlin/au/com/dius/pact/provider/junit5/PactJUnit5VerificationProvider.kt
oswaldquek: I would but `au.com.dius.pact.provider.junit.loader.PactSource` has already been imported.

provider/pact-jvm-provider/src/main/groovy/au/com/dius/pact/provider/ProviderVerifier.groovy
oswaldquek: I'm not sure myself. I copied it from https://github.com/DiUS/pact-jvm/blob/master/provider/pact-jvm-provider/src/main/groovy/au/com/dius/pact/provider/reporters/JsonReporter.groovy#L16

uglyog: Hmm, I think IntelliJ must have slipped that one in. But it works ok, so all good.

provider/pact-jvm-provider/src/main/kotlin/au/com/dius/pact/provider/reporters/AnsiConsoleReporter.kt
oswaldquek: fixed

PR #923, opened 2019-08-20
Extended Kotlin DSL

gmariotti:
The purpose of this pull request is to improve the Kotlin experience when using the consumer DSL. Differently from the previous `LambdaDslJsonArray.newObject`, the new methods avoid the repetition of `it` inside the lambda by using, as a parameter, a lambda with receiver.

Questions:
- What is the best way to test these extensions?
- What kind of documentation is it needed (JavaDoc, README examples, ...)

Thank you in advance for investing your time in looking into it.

### DIFF
--- a/consumer/pact-jvm-consumer-java8/README.md
+++ b/consumer/pact-jvm-consumer-java8/README.md
@@ -189,15 +189,22 @@ newJsonArray((rootArray) -> {
     rootArray.array((a) -> a.numberValue(1).numberValue(2));
     rootArray.array((a) -> a.object((o) -> o.stringValue("foo", "Foo")));
 }).build();
-
 ```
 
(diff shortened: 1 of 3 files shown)


uglyog:
Thanks, this looks useful. To test it further than what you did is to write a consumer test, maybe in pact-jvm-consumer-junit or pact-jvm-consumer-junit5. But that's not too important, your test shows how it works.

For documentation, just add a section to the README with an example of how to use it with Kotlin.

gmariotti:
Hi @uglyog, I've updated the pull request with some documentation and an example in the DSL README. I've also added a couple of extension methods for `LambdaDslJsonArray`, however, I had to change the signature of `LambdaDslJsonArray.newObject` so that `this` can be used instead of `it`. Is this a problem?

uglyog:
All good. Thanks!

gmariotti:
Thank you for accepting it!!

Issue #929, opened 2019-08-29
pact-jvm-provider-gradle uses project.version for both pactVerify and pactPublish tasks.


toastyblast:
When I set
```
pact {
   serviceProviders {
      'Provider-Name' {
         // Mock server host settings...

         // This is changing the project.version, not just the provider version for pact!
         version = branchName() + '-' + abbreviatedId()

         if ('pactVerify' in gradle.startParameter.taskNames) {
            hasPactsFromPactBroker('http://localhost:80/')
         }
      }
   }
}
```
for pactVerify or for pactPublish
```
pact {
   serviceProviders {
        // Again, this is changing the project.version, not just the consumer version for pact!
        version = 'PROV-TEST-' + grgit.head().abbreviatedId
        tags = [grgit.branch.current.name]

        pactDirectory = file('./test/pact/consumerPacts')
        pactBrokerUrl = 'http://localhost:80/'
   }
}
```
The value setting version in either one of them also changes `project.version`. In our team we want the consumer and provider version to be the git name and hash of the branch the pact as been sent/verified from to the broker for clarity, and not the Gradle project version. This is currently causing a breaking issue, as the jar files we generate DO use the project.version, and we do not want those jar files to change names!

Can this please be changed/fixed to allow one to set provider and consumer versions different from project.version? I understand it as a _default_, but if someone wants it changed, they shouldn't have to change the _whole_ Gradle project's version.

We're using pact-jvm-provider-gradle-2.12 version 3.6.12.

toastyblast:
My excuses, the second one is incorrect. Setting version in the publish section does not change the project.version. That only makes me more surprised though, why do that and not allow the same for verify?

uglyog:
Response is on the PR #931

toastyblast:
Thank you for the heads up. I've seen the release, I'll get to implementing it for my team now. Once again, many thanks for the help!

PR #931, opened 2019-08-30
Provider gradle add verification provider version

toastyblast:
[Issue #929](https://github.com/DiUS/pact-jvm/issues/929) - Added "version" variable to Pact Provider during the pactVerify task in pact-jvm-provider-gradle.
This is to stop pact from using the project.version as its hardcoded value (it's now the default, if no version is given to the serviceProvider).


toastyblast:
I'm not fully sure if my change in the PR above actually does what I want it to, as I'm not familiar with the codebase (and I can't test it in the project I need it in until it is released in a version I can add as a dependency in Gradle). But if it indeed doesn't work, I at least hope it gets the idea across of what I need.

toastyblast:
Also may I add that when working on this, I couldn't run the full ./gradlew clean build command as suggested in the "How to contribute" page of pact. It will fail on task "run 200 requests to mock server". Any way to fix this so I can run builds properly locally?

Additionally, when trying to do task "signMavenProductionProduction" it will also fail, saying there's no signatory. Any suggestions for this as well?

gradlew.bat
```diff
@@ -14,7 +30,7 @@ set APP_BASE_NAME=%~n0
 set APP_HOME=%DIRNAME%
 
 @rem Add default JVM options here. You can also use JAVA_OPTS and GRADLE_OPTS to pass JVM options to this script.
-set DEFAULT_JVM_OPTS=
+set DEFAULT_JVM_OPTS="-Xmx64m" "-Xms64m"
```
Scot3004: This environment value is put by the CI server, could you please remove it

gradlew
```diff
@@ -28,7 +44,7 @@ APP_NAME="Gradle"
 APP_BASE_NAME=`basename "$0"`
 
 # Add default JVM options here. You can also use JAVA_OPTS and GRADLE_OPTS to pass JVM options to this script.
-DEFAULT_JVM_OPTS=""
+DEFAULT_JVM_OPTS='"-Xmx64m" "-Xms64m"'
```
Scot3004: This environment value is put by the CI server, could you please remove it

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,5 +1,5 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-5.4.1-all.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-5.4.1-bin.zip
```
Scot3004: Why did you changed this value?

gradlew
```diff
@@ -1,5 +1,21 @@
 #!/usr/bin/env sh
 
+#
```
Scot3004: i don't think that you should update this file, i not find relationship between your issue and this file

toastyblast: All issues you mentioned weirdly I haven't changed manually at all. Maybe my IntelliJ set them automatically, or Gradle did. Wouldn't be able to tell you, but I'll remove/change everything you mentioned 👍

toastyblast:
Problem is, when I do "./gradlew clean build" (as instructed by the contribution guide), it still fails 
":consumer:pact-jvm-consumer:test > Executing test au.com.dius.pact.consumer.MockHttpServerSpec > handle more than 200 tests"
due to a "org.spockframework.runtime.SpockTimeoutError at MockHttpServerSpec.groovy:57".

uglyog:
That test has a 60 second timeout on it. It's probably better to have that run on CI only. I'll make the change.

Don't know why you're getting the signMavenProductionProduction error. It should only require a key to sign the JARs when publishing. You can run run `./gradlew build x signMavenProductionProduction` to skip it.

uglyog:
After reviewing this PR, I think the better way is to allow you to override the callback that returns the version instead of introducing a new property.

I.e.,

```
pact {
   serviceProviders {
      'Provider-Name' {
         // Mock server host settings...

         providerVersion = { branchName() + '-' + abbreviatedId() }

         if ('pactVerify' in gradle.startParameter.taskNames) {
            hasPactsFromPactBroker('http://localhost:80/')
         }
      }
   }
}
```

toastyblast:
So, does this already work, or is this just a proposal? I fear I'm not well acquainted enough with the workflow or code of Pact to implement such a thing myself at the moment.

uglyog:
After reviewing the code, the callback exists but it set by the Task, not in the configuration. But I'm going to move the property to the Gradle specific class so it won't effect any other build tools.

I'm also thinking of renaming the publish property to providerVersion so they can't be confused with the project version from Gradle.

toastyblast:
So, after this change was merged a new version was created, however, this version does not include the fix.

Now 15 days later, I still haven't seen any updates. Can I be given any idea or indication of when to expect this change to be implemented so we can make use of it?

uglyog:
It's merged and will be in the next release. I'll organise a release this weekend.

toastyblast:
Thank you for the heads up. I've seen the release, I'll get to implementing it for my team now. Once again, thanks a lot for the help!

Issue #925, opened 2019-08-25
Support bearer tokens with JUnit annotations


uglyog:
Currently bearer tokens are supported with the JUnit annotations by using the basic auth annotation, setting the username to the token, the scheme to bearer and the password to an empty string.

i.e.

```java
@Provider("Animal Profile Service")
@PactBroker(host = "pactbroker", scheme = "https",
  authentication = @PactBrokerAuth(scheme = "bearer", username = "<TOKEN VALUE>", password = ""))
public class PactJUnitBrokerTest {
```

It would be desirable to have an explicit token value, by either doing

```java
@Provider("Animal Profile Service")
@PactBroker(host = "pactbroker", scheme = "https",
  authentication = @PactBrokerAuth(token = "<TOKEN VALUE>"))
public class PactJUnitBrokerTest {
```

or something like

```java
@Provider("Animal Profile Service")
@PactBroker(host = "pactbroker", scheme = "https",
  authentication = @BearerToken("<TOKEN VALUE>"))
public class PactJUnitBrokerTest {
```

PR #951, opened 2019-10-05
Support bearer token with JUnit annotations - Fixes gh-925

tinexw:
I decided to do the first option since it was easier to implement. Should I add checks somewhere that the correct combination of values is set?

### DIFF
--- a/provider/pact-jvm-provider-junit/README.md
+++ b/provider/pact-jvm-provider-junit/README.md
@@ -212,7 +212,6 @@ The following keys may be managed through the environment
 * `pactbroker.port`
 * `pactbroker.scheme`
 * `pactbroker.tags` (comma separated)
-* `pactbroker.auth.scheme` (basic or bearer)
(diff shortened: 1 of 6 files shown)


uglyog:
I think the scheme can be defaulted if the token property is set. That way people will only need to set the token value. I.e.,

```java

**Comment by tinexw** [comment]
Yes, makes sense. Will add this.

**Comment by tinexw** [comment]
I changed the implementation so that it is now possible to say

```
authentication = @PactBrokerAuth(token = "test")
```

I marked the schema as deprecated. It is now automatically set to
- `Basic` if the username is set
- `Bearer` if the token is set

I  hope marking it as deprecated is ok. It did not see it being used anywhere. Please let me know if it is actually of use somewhere and we should keep it.

uglyog:
Deprecated makes sense. It's really there in case other authentication schemes are required, but the implementation would not be supported so it is redundant.

PR #955, opened 2019-10-12
:sparkles: Support Bearer Authentication for verifying pacts with the maven plugin

### LINKED ISSUE #954: Support Bearer Authentication for Pact Verification with Maven plugin
# Background
Currently, the Maven plugin only supports Basic Authentication with Pact Brokers for verifying pacts, although it does support Bearer Authentication for publishing pacts. Given that https://pactflow.io uses Bearer Authentication, I think it makes sense to add this functionality to the `PactProviderMojo`. 

# Proposed Change
I propose the following:

1. Rename the `BasicAuth` class to simply be `PactBrokerAuth`. It will have the fields `username` and `password` like it has today but it will have another field that is called `scheme`. Supported values for this field will be `"basic"` and `"bearer"` with a default value of `"basic"`
2. Modify the snippet of code handling authentication configuration in the `PactProviderMojo` to be the following:
```kotlin
 if (pactBroker?.authentication != null) {
      if (provider.pactBroker.authentication.password != null) {
        options["authentication"] = listOf(provider.pactBroker.authentication.scheme, provider.pactBroker.authentication.username,
                provider.pactBroker.authentication.password)
      } else {
        options["authentication"] = listOf(provider.pactBroker.authentication.scheme, provider.pactBroker.authentication.username)
      }
    }
```

I suggested this approach because it feels in spirit with how some of the other pact broker integrations work, as enumerated here: https://docs.pactflow.io/docs/troubleshooting/

# Testing
I think an added test case to `PactProviderMojoSpec` should suffice, but open to debate here.

ryandens:
- Modified the `BasicAuth` class to encapsulate the authentication mechanism for verifying pacts from pact brokers. This includes an extra field on the class called `scheme`, which has a default value of `basic` so as not to interfere with existing users of basic auth
- Modified the setting of the `authentication` option to only include `PactBrokerAuth#password` if password is not `null` and to set the scheme from config rather than a hardcoded value.
- Update docs to demonstrate how to use basic and bearer authentication for verifying pacts
- Added unit test for verifying authentication header gets set correctly if using bearer authentication

closes #954

### DIFF
--- a/provider/pact-jvm-provider-maven/README.md
+++ b/provider/pact-jvm-provider-maven/README.md
@@ -366,16 +366,16 @@ For example:
 
 ### Verifying pacts from an authenticated pact broker
 
-If your pact broker requires authentication (basic authentication is only supported), you can configure the username
+If your pact broker requires authentication (basic and bearer authentication are supported), you can configure the username
 and password to use by configuring the `authentication` element of the `pactBroker` element of your provider.
 
-For example:
+For example, here is how you configure the plugin to use basic authentication for verifying pacts:
 
 ```xml
 <plugin>
     <groupId>au.com.dius</groupId>
     <artifactId>pact-jvm-provider-maven</artifactId>
-    <version>4.0.0</version>
+    <version>4.0.1</version>
     <configuration>
       <serviceProviders>
         <serviceProvider>
@@ -384,8 +384,34 @@ For example:
           <pactBroker>
               <url>http://pactbroker:1234</url>
               <authentication>
+                  <scheme>basic</scheme>
                   <username>test</username>
                   <password>test</password>
+              </authentication>
+          </pactBroker>
+        </serviceProvider>
+      </serviceProviders>
+    </configuration>
+</plugin>
+```
+
+Here is how you configure the plugin to use basic authentication for verifying pacts
+
+```xml
+<plugin>
+    <groupId>au.com.dius</groupId>
+    <artifactId>pact-jvm-provider-maven</artifactId>
+    <version>4.0.1</version>
+    <configuration>
+      <serviceProviders>
+        <serviceProvider>
+          <name>provider1</name>
+          <stateChangeUrl>http://localhost:8080/tasks/pactStateChange</stateChangeUrl>
+          <pactBroker>
+              <url>http://pactbroker:1234</url>
+              <authentication>
+                  <scheme>bearer</scheme>
+                  <username>TOKEN</username>
               </authentication>
           </pactBroker>
         </serviceProvider>
--- a/provider/pact-jvm-provider-maven/README.md
+++ b/provider/pact-jvm-provider-maven/README.md
@@ -395,7 +395,7 @@ For example, here is how you configure the plugin to use basic authentication fo
 </plugin>
 ```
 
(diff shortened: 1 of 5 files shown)


ryandens:
I also ran a basic manual test with my project to verify that the maven plugin can now pull pacts from and publish verification results to Pactflow

provider/pact-jvm-provider-maven/README.md
```diff
@@ -394,6 +395,31 @@ For example:
 </plugin>
 ```
 
+Here is how you configure the plugin to use basic authentication for verifying pacts
```
uglyog: This should be: Here is how you configure the plugin to use bearer token authentication for verifying pacts

provider/pact-jvm-provider-maven/README.md
```diff
@@ -394,6 +395,31 @@ For example:
...
+    <configuration>
+      <serviceProviders>
+        <serviceProvider>
+          <name>provider1</name>
+          <stateChangeUrl>http://localhost:8080/tasks/pactStateChange</stateChangeUrl>
+          <pactBroker>
+              <url>http://pactbroker:1234</url>
+              <authentication>
+                  <scheme>bearer</scheme>
+                  <username>TOKEN</username>
```
uglyog: instead of re-using the `username` attribute, it would be better to introduce a `token` attribute

provider/pact-jvm-provider-maven/src/main/kotlin/au/com/dius/pact/provider/maven/PactProviderMojo.kt
```diff
@@ -136,8 +136,12 @@ open class PactProviderMojo : AbstractMojo() {
     val options = mutableMapOf<String, Any>()
 
     if (pactBroker?.authentication != null) {
-      options["authentication"] = listOf("basic", provider.pactBroker.authentication.username,
-        provider.pactBroker.authentication.password)
+      if (provider.pactBroker.authentication.password != null) {
```
uglyog: This logic should be based on the scheme. If someone sets the scheme to 'bearer', but then sets the password to a non-empty value, this will not work.

Better to have a token attribute, and we can default the scheme to 'bearer' if that is set.

provider/pact-jvm-provider-maven/README.md
ryandens: 👍 thanks!

provider/pact-jvm-provider-maven/README.md
ryandens: done! I re-used the `username` attribute because that's how the junit extensions appear to work based on some documentation https://docs.pactflow.io/docs/troubleshooting/ but I see that other integrations don't work this way

provider/pact-jvm-provider-maven/src/main/kotlin/au/com/dius/pact/provider/maven/PactProviderMojo.kt
ryandens: done, and added a unit test to assert the new behavior

uglyog:
Looks good, thanks!

uglyog:
4.0.1 has been released

ryandens:
Thank you!

Issue #976, opened 2019-11-05
Category.copyWithUpdatedMatcherRootPrefix doesn't update if root prefix string is prepended with the prefix already


GregPappasWahanda:
When adding a PactDslJsonBody to a parent through the use of PactDslJsonBody.object
we can experience an issue where if the parent body is called `blue` (for example)

and the new part being added is called `blueberry` then the key is not updated to be `blue.blueberry`
and we see: 
```
    "$.fruitByColour.blueberry": {
         "matchers": [
              {
                  "match": "type",
                }
              ],
              "combine": "AND"
            },
"fruitByColour": {
            "blue": {
              "blueberry": {
               <-- some random blueberry properties here -->
              },
```
this happens because of:
```
PactDslJsonBody().object(String name, DslPart value)
|
|_new PactDslJsonBody(base, "", this, (PactDslJsonBody) value) 
   |
   |_ Category.copyWithUpdatedMatcherRootPrefix
      |
      |_  e.key.startsWith(prefix) -> e.key
```

PullRequest is here: 
https://github.com/DiUS/pact-jvm/pull/977

uglyog:
4.0.3 has been released with this fix

GregPappasWahanda:
Yep looks good thank you!

PR #977, opened 2019-11-05
#976: always prepend prefix to existing key

GregPappasWahanda:
When adding a PactDslJsonBody to a parent through the use of PactDslJsonBody.object
we can experience an issue where if the parent body is called

blue (for example)
and the new part being added is called blueberry then the key is not updated to be blue.blueberry
and we see:
```
    "$.fruitByColour.blueberry": {
         "matchers": [
              {
                  "match": "type",
                }
              ],
              "combine": "AND"
            },
"fruitByColour": {
            "blue": {
              "blueberry": {
               <-- some random blueberry properties here -->
              },
```
this happens because of:
```
PactDslJsonBody().object(String name, DslPart value)
|
|_new PactDslJsonBody(base, "", this, (PactDslJsonBody) value) 
   |
   |_ Category.copyWithUpdatedMatcherRootPrefix
      |
      |_  e.key.startsWith(prefix) -> e.key
```
https://github.com/DiUS/pact-jvm/issues/976


uglyog:
Thanks for the PR

PR #985, opened 2019-11-25
fixes 953 by upgrading ktor to a version that uses kotlin 1.3.

scheuchzer:


scheuchzer:
Build failed because it failed to download the scala-compiler.jar in two cases.

PR #999, opened 2020-01-19
feat: Include tag in jUnit test name

aplsup:
In a scenario where pact provider verification tests possibly run against multiple tags, it would be a nice to have, in case of failures, to easily identify which tag (if present) has caused the failure.

Example of jUnit output:
```
Test suite name: 
      [tag:#tag-name] #consumer-name: Upon #interaction-name
```

### DIFF
--- a/provider/pact-jvm-provider-junit/src/test/groovy/au/com/dius/pact/provider/junit/descriptions/DescriptionGeneratorTest.groovy
+++ b/provider/pact-jvm-provider-junit/src/test/groovy/au/com/dius/pact/provider/junit/descriptions/DescriptionGeneratorTest.groovy
@@ -34,7 +34,7 @@ class DescriptionGeneratorTest extends Specification {
   def 'when BrokerUrlSource tests description includes tag if present'() {
     def interaction = new RequestResponseInteraction('Interaction 1',
             [ new ProviderState('Test State') ], new Request(), new Response())
-    def pact = new RequestResponsePact(new Provider(), new Consumer(), [ interaction ])
+    def pact = new RequestResponsePact(new Provider(), new Consumer("the-consumer-name"), [ interaction ])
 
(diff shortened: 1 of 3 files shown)


provider/pact-jvm-provider-junit/src/test/groovy/au/com/dius/pact/provider/junit/descriptions/DescriptionGeneratorTest.groovy
```diff
@@ -0,0 +1,62 @@
...
+    def pact = new RequestResponsePact(new Provider(), new Consumer(), [ interaction ])
+
+    expect:
+    def pactSource =  new BrokerUrlSource('url', 'url', [:], [:], tag)
+    def generator = new DescriptionGenerator(clazz, pact, pactSource)
+    description == generator.generate(interaction).methodName
+
+    where:
+    tag      | description
+    'master' | '[tag:master] consumer - Upon Interaction 1'
```
bethesque: Is 'consumer' the name of the consumer? To me, it would make more sense if the `[tag:master]` came after the consumer name - but that's just a very small preference.

aplsup: I personally agree, appreciate the suggestion. Will apply change

bethesque:
I ain't no expert, but LGTM.

bethesque:
Can you have a look at the failing PR builds @aplsup? https://travis-ci.org/DiUS/pact-jvm/builds/640724361 For some reason code narc is being narcy on a few of the JVMs.

aplsup:
@bethesque fixed here https://github.com/DiUS/pact-jvm/pull/1003

PR #1012, opened 2020-01-31
The request body should be encoded according to the request header

BenVercammen:
Currently the setupBody() method in the ProviderClient class does not set a contentType. This results in the StringEntity() constructor to use the HTTP.DEF_CONTENT_CHARSET, which in turn resolves to Consts.ISO_8859_1.

This is a problem when you have a UTF-8 encoded pact.json file with special characters. The request that is being sent by the mock consumer, will read the UTF-8 encoded body (from the pact file) as ISO-8859-1, thus messing up any special characters it contains.

I've also got a very basic Spring Boot project set up to show the problem at https://github.com/BenVercammen/pact-jvm-provider-encoding-issue.git
(make sure the pact files are properly encoded, probably will check out in default encoding)

Disclaimer: I'm not familiar with kotlin or groovy...


uglyog:
Thanks for the PR!

uglyog:

PR #1016, opened 2020-02-04
feat: Use Before/AfterTestExecutionCallback

asteffey:
## Problem to solve

The junit5 PactConsumerTestExt currently runs [before](https://github.com/DiUS/pact-jvm/blob/3860faa5d3e67b7bbf42d25cc7e689314a905859/consumer/pact-jvm-consumer-junit5/src/main/kotlin/au/com/dius/pact/consumer/junit5/PactConsumerTestExt.kt#L192) and [after](https://github.com/DiUS/pact-jvm/blob/3860faa5d3e67b7bbf42d25cc7e689314a905859/consumer/pact-jvm-consumer-junit5/src/main/kotlin/au/com/dius/pact/consumer/junit5/PactConsumerTestExt.kt#L298) all user-defined `@BeforeEach` and `@AfterEach` annotated methods in the Pact consumer tests, respectively. This is a problem when a user-defined `@BeforeEach` method needs to setup some test state before a `@Pact` annotated builder method is called.

It is more natural for PactConsumerTestExt to be run after any user-defined `@BeforeEach` and before any user-defined `@AfterEach` methods.

## Proposed changes

Use `BeforeTestExecutionCallback` and `AfterTestExecutionCallback` is used instead of `BeforeEachCallback` and `AfterEachCallback`, respectively.

## NOTICE


uglyog:

Issue #916, opened 2019-07-29
pact.consumer.junit5.PactConsumerTestExt WARNING: An illegal reflective access operation has occurred


khillyard:
Running a test method yields warning:

_20:55:40.880 [main] DEBUG au.com.dius.pact.consumer.junit5.PactConsumerTestExt - Invoking method 'locationMatch' to get Pact for the test 'locationMatchSucceeds'
WARNING: An illegal reflective access operation has occurred
WARNING: Illegal reflective access by org.codehaus.groovy.vmplugin.v7.Java7$1 (file:/Users/Joe/.m2/repository/org/codehaus/groovy/groovy/2.5.7/groovy-2.5.7.jar) to constructor java.lang.invoke.MethodHandles$Lookup(java.lang.Class,int)
WARNING: Please consider reporting this to the maintainers of org.codehaus.groovy.vmplugin.v7.Java7$1
WARNING: Use --illegal-access=warn to enable warnings of further illegal reflective access operations
WARNING: All illegal access operations will be denied in a future release_

Should I report this to org.codehaus.groovy or is this a version dependency issue?

uglyog:
Groovy 3.0 is going to support JDK 9+, but it is still in beta. Although this warning is displayed, I haven't noticed anything failing due to this. So you can ignore it and I'll update the Groovy version to 3.0 when it is out of beta.

cogman:
Groovy 3 is (now?) released.  We've seen with our testing of Java 14 RC that this warning becomes an error.

cogman:
I tried to do a bump, however, it looks like Spock 1.3 doesn't support groovy 3.

This looks like the root issue.

https://github.com/spockframework/spock/issues/997

Also, my build breaks in scala.  Is this expected?  I wasn't able to get a clean compile with no changes.  I saw this instead

```
pact-jvm\pact-jvm-server\src\main\scala\au\com\dius\pact\server\Complete.scala:13: non-variable type argument AnyRef in type pattern scala.collection.immutable.Map[AnyRef,AnyRef] (the underlying of Map[AnyRef,AnyRef]) is unchecked since it is eliminated by erasure
    case map: Map[AnyRef, AnyRef] => {
```

cogman:
Looks like the changes are more fun than I thought.  http-builder does not work with groovy 3.  I tried moving to http-builder-ng, but looks like more changes are needed to make that work.

uglyog:
I've been waiting for Spock to support Groovy 3.

[...]

PR #1019, opened 2020-02-11
fix: 916 use groovy 3

cogman:
This still fails some ABC code narc tests.  Does anyone have a suggestion on how to resolve this?

Fixes issue- #916

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -76,7 +76,7 @@ subprojects {
       testCompile "org.junit.jupiter:junit-jupiter-api:${project.junit5Version}"
       testRuntime "org.junit.jupiter:junit-jupiter-engine:${project.junit5Version}"
       testCompile 'org.hamcrest:hamcrest:2.1'
-      testCompile('org.spockframework:spock-core:1.3-groovy-2.5') {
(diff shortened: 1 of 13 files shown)


uglyog:
ABC normally indicates methods that are too complex, but maybe that is CodeNarc not dealing with Groovy 3. I can have a look into fixing that.

cogman:
@uglyog  Thanks for looking into this.  It may just be that the switch to http-builder-ng pushed it over the edge.  It changed the parameters from maps to more of a DSL which means more assignments.

I didn't want to do a whole lot of restructuring of tests so I backed off with just the narcs I could simply solve.  As is it should be pretty close to the same flow it was before my tinkering.

uglyog:
I've fixed the ABC metric violation, as well as some other tests that still were using the old HTTPBuilder.

The next big problem is the Gradle plugin. This has to be compiled against the version of Groovy included with Gradle, which is currently 2.5.4. This causes an issue with both Groovy and Spock in that module.

uglyog:
Green build!

cogman:
Thanks for taking this past the finish line!

uglyog:
4.0.6 has been released

Issue #1020, opened 2020-02-17
SpringRestPactRunner instantiates ApplicationContext too eagerly


dbevacqua:
`SpringRestPactRunner` instantiates an ApplicationContext in `getPactSource()`, which is called in `PactRunner`'s constructor. When run with Surefire (and probably other tools), the logic to filter tests by `@Category` requires instantiation of `Runner` implementations (see e.g. https://github.com/apache/maven-surefire/blob/surefire-2.22.2/surefire-providers/surefire-junit47/src/main/java/org/apache/maven/surefire/junitcore/JUnitCoreWrapper.java#L129).

Therefore, even if a test class which uses `SpringRestPactRunner` is configured to not to run, an `ApplicationContext` will be created - a potentially expensive and time-consuming operation.

An obvious fix is to defer initialisation of `PactRunner` until it is actually required, i.e. when the `Runner` is `run()`.

mikahjc:
`ApplicationContext` is generated for a `SpringEnvironmentResolver` while loading pacts, but I can only see that `PactLoader`'s `ValueResolver` is used in `PactBrokerLoader`. I don't see how this could be deferred unless we had some sort of placeholder test that can load the pacts when the tests actually get run. But I'm not sure how I feel about that idea.

PR #1021, opened 2020-02-17
Defer Runner initialization until it is run

dbevacqua:
Fixes https://github.com/DiUS/pact-jvm/issues/1020


uglyog:

PR #1024, opened 2020-02-24
Extended PactFilter to allow variable filter criterias

Pfarrer:
Add a filter property to the PactFilter annotation to allow generic filtering, not only based on the provider state. The implementation is a non breaking change since the default behaviour is just like before but it not also allows to filter for the request path.

What do you think about this?

### DIFF
--- a/consumer/pact-jvm-consumer/src/test/java/au/com/dius/pact/consumer/PactDslJsonBodyTest.java
+++ b/consumer/pact-jvm-consumer/src/test/java/au/com/dius/pact/consumer/PactDslJsonBodyTest.java
@@ -265,7 +265,7 @@ public class PactDslJsonBodyTest {
         .date("lastUpdate", DATE_FORMAT)
         .date("creationDate", DATE_FORMAT);
       JSONObject jsonObject = (JSONObject) response.getBody();
-      assertThat(jsonObject.get("lastUpdate").toString(), matchesPattern("\\w{3}\\.?, \\d{2} \\w{3}\\.? \\d{4} \\d{2}:00:00 \\+\\d+ GMT"));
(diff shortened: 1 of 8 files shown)


consumer/pact-jvm-consumer/src/test/java/au/com/dius/pact/consumer/PactDslJsonBodyTest.java
```diff
@@ -265,7 +265,7 @@ public void testLargeDateFormat() {
         .date("lastUpdate", DATE_FORMAT)
         .date("creationDate", DATE_FORMAT);
       JSONObject jsonObject = (JSONObject) response.getBody();
-      assertThat(jsonObject.get("lastUpdate").toString(), matchesPattern("\\w{3}\\.?, \\d{2} \\w{3}\\.? \\d{4} \\d{2}:00:00 \\+\\d+ GMT"));
+      assertThat(jsonObject.get("lastUpdate").toString(), matchesPattern("\\w{2,3}\\.?, \\d{2} \\w{3}\\.? \\d{4} \\d{2}:00:00 \\+\\d+ GMT"));
```
Pfarrer: I fixed this test that is not related to my PR. The day field was 'Mo' for me, not sure why, maybe be cause of my locale?
In German, 'Mo' is short for Montag which is Monday. :)

Pfarrer: I fixed this test that is not related to my PR. The day field was 'Mo' for me, not sure why, maybe be cause of my locale?
In German, 'Mo' is short for Montag which is Monday. :)

Pfarrer: I fixed this test that is not related to my PR. The day field was 'Mo' for me, not sure why, maybe be cause of my locale?
In German, 'Mo' is short for Montag which is Monday. :)

Pfarrer:
@rholshausen For some reason, I cannot add reviewers.
Can you please have a look at this?

PR #1030, opened 2020-02-27
Avoid unhandled exception in `validateTestTarget`

kppullin:
`InteractionRunner.validateTestTarget()` throws an `IndexOutOfBoundsException`
when it tries to `return annotatedFields[0]` if the collection is empty.

Instead it should return nothing/void as it adds an `Exception` to the
`errors` list prior to the return statement, and this fulfills the
expectation of the caller and mirrors the other `validateX()` methods
in the class.


uglyog:
Thanks for the PR

PR #1034, opened 2020-02-28
Fix off-by-one valueFromProviderState index

kppullin:
`PactDslJsonArray.valueFromProviderState()` assigns an index
one less than it should.

Before appending to the `body` array,
the method invokes `appendArrayIndex()`, which in turn emits an
index value of `body.length()-1`. Since the `body` array is yet to
be updated the calculated index is off by one (the most obvious
sign of this is a negative index of `-1`).

This PR moves the append to `body` first.

### DIFF
--- a/consumer/pact-jvm-consumer/src/main/java/au/com/dius/pact/consumer/dsl/PactDslJsonArray.java
+++ b/consumer/pact-jvm-consumer/src/main/java/au/com/dius/pact/consumer/dsl/PactDslJsonArray.java
@@ -1232,8 +1232,8 @@ public class PactDslJsonArray extends DslPart {
    * @param example Example value to be used in the consumer test
    */
   public PactDslJsonArray valueFromProviderState(String expression, Object example) {
-    generators.addGenerator(Category.BODY, rootPath + appendArrayIndex(0), new ProviderStateGenerator(expression));
     body.put(example);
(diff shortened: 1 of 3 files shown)


consumer/pact-jvm-consumer/src/main/java/au/com/dius/pact/consumer/dsl/PactDslJsonArray.java
```diff
@@ -1232,8 +1232,8 @@ public PactDslJsonArray eachArrayWithMinMaxLike(int numberExamples, Integer minS
    * @param example Example value to be used in the consumer test
    */
   public PactDslJsonArray valueFromProviderState(String expression, Object example) {
-    generators.addGenerator(Category.BODY, rootPath + appendArrayIndex(0), new ProviderStateGenerator(expression));
     body.put(example);
+    generators.addGenerator(Category.BODY, rootPath + appendArrayIndex(0), new ProviderStateGenerator(expression));
```
kppullin: Another option here would be to change `appendArrayIndex(0)` to `appendArrayIndex(1)`.  I went the proposed approach instead as it matched most other patterns in the class and it didn't _appear_ there was any dependencies on the order of operations here.

uglyog:
Thanks for the PR! I agree with your approach being the better option.

uglyog:
Version 4.0.7 has been released with this change

PR #1036, opened 2020-02-28
Revert #1021 Defer JUnit runner initialization

### LINKED ISSUE #1035: Deferred PactRunner initialization causes JUnit 4.13 to not pick up any pacts
The changes from #1021 are causing the test runner to not have any children when `getFilteredChildren()` in JUnit is called when the runner is configured by JUnit. `getFilteredChildren()` does a one-time initialization of the `filteredChildren` field, but gets called before `PactRunner.run()` does. I'm seeing this with the Maven Surefire plugin and with IntelliJ's JUnit runner on JUnit 4.13. JUnit 4.12 doesn't exhibit this behavior because of a change in the `getFilteredChildren()` method.

One solution is to call `initialize()` when `getChildren()` is called, but that may counteract the changes in #1020/#1021.

mikahjc:
Fixes #1035. @dbevacqua, could you see if this causes a regression of #1020? If so, we should rethink how initialization is handled.

### DIFF
--- a/provider/pact-jvm-provider-junit/src/main/kotlin/au/com/dius/pact/provider/junit/PactRunner.kt
+++ b/provider/pact-jvm-provider-junit/src/main/kotlin/au/com/dius/pact/provider/junit/PactRunner.kt
@@ -53,20 +53,8 @@ open class PactRunner<I>(clazz: Class<*>) : ParentRunner<InteractionRunner<I>>(c
 
   private val child = mutableListOf<InteractionRunner<I>>()
   private var valueResolver = SystemPropertyResolver()
-  private val clazz = clazz
-  private var initialized = false
 
-  override fun run(notifier: RunNotifier?) {
-    initialize()
-
-    super.run(notifier)
-  }
-
-  private fun initialize() {
-    if (initialized) {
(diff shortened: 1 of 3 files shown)


dbevacqua:
Unfortunately I think this will indeed cause a regression. It is going to be tricky to resolve. Initialising the children involves creating new `InteractionRunner`s, which require a `PactSource`. The `PactSource` for a `SpringRestPactRunner` requires a `ValueResolver`, which in this case requires an `Environment` from an `ApplicationContext` (see https://github.com/DiUS/pact-jvm/blob/4_0_6/provider/pact-jvm-provider-spring/src/main/kotlin/au/com/dius/pact/provider/spring/SpringRestPactRunner.kt#L57-L59). I have not yet determined where the `ValueResolver` is used but it is not obvious whether its initialization can be deferred.

Given that #1020 is not a bug as such, and that `@Category` is an experimental feature of JUnit 4 which has been superseded by `@Tag` with filters (which may or may not suffer from #1020 - I have not tested this), we should probably revert #1021 until/unless we can see a way to defer initialization of the `ValueResolver`

PR #1037, opened 2020-03-02
Revert "Defer Runner initialization until it is run"

### LINKED ISSUE #1035: Deferred PactRunner initialization causes JUnit 4.13 to not pick up any pacts

mikahjc:
Fixes #1035. Per discussion on #1036, it seems to be best to revert #1021 until/unless a better way to defer runner initialization can be found.


uglyog:
I'd prefer not to just revert that change. Having the construction of the pact tests in the constructor has always been a smell. I think it is better having it in the separate method, even if we then just call that from the constructor to restore the previous behaviour.

I'm going to close this PR, but will now investigate the issue to see if we can find a better solution.

PR #1042, opened 2020-03-05
Perform same check on character validity in pathIdentifier that identifier does

### LINKED ISSUE #1041: pathIdentifier doesn't recognize stringified negative numbers as valid
During an upgrade from 3.6.14 to 4.0.6, we saw that a pact test started failing on the consumer side because the response JSON has keys like `-100`, `-200`, etc. These are valid JSON keys and JSONPath specifiers, but [`pathIdentifier`](https://github.com/DiUS/pact-jvm/blob/3bc07b834f651f1aa17663416786cdae7a9ce0d6/core/model/src/main/kotlin/au/com/dius/pact/core/model/PathExpressions.kt#L95) in `PathExpressions` doesn't recognize `-` as a valid character, nor does it perform the same validation that [`identifier`](https://github.com/DiUS/pact-jvm/blob/3bc07b834f651f1aa17663416786cdae7a9ce0d6/core/model/src/main/kotlin/au/com/dius/pact/core/model/PathExpressions.kt#L71) does.

mikahjc:
Fixes #1041

### DIFF
--- a/core/model/src/main/kotlin/au/com/dius/pact/core/model/PathExpressions.kt
+++ b/core/model/src/main/kotlin/au/com/dius/pact/core/model/PathExpressions.kt
@@ -92,7 +92,7 @@ fun pathIdentifier(
     val ch = chars.next()
     when {
       ch.value == '*' -> tokens.add(PathToken.Star)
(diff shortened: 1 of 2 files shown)


uglyog:
Thanks for the PR

Issue #1013, opened 2020-02-04
MalformedJsonException is thrown is thrown if a response does not have Content-Type header


flexfrank:
I test a API that does not have Content-Type header by pact.
ProviderClient treats its response body as `application/json` and failed to parse.
I think the should be treated as `application/octet-stream` by default.

```
0 - com.google.gson.stream.MalformedJsonException: Use JsonReader.setLenient(true) to accept malformed JSON at line 1 column 7 path $
com.google.gson.stream.MalformedJsonException: Use JsonReader.setLenient(true) to accept malformed JSON at line 1 column 7 path $
	at com.google.gson.stream.JsonReader.syntaxError(JsonReader.java:1564)
	at com.google.gson.stream.JsonReader.checkLenient(JsonReader.java:1405)
	at com.google.gson.stream.JsonReader.doPeek(JsonReader.java:543)
	at com.google.gson.stream.JsonReader.peek(JsonReader.java:426)
	at com.google.gson.JsonParser.parseReader(JsonParser.java:61)
	at com.google.gson.JsonParser.parseString(JsonParser.java:47)
	at com.google.gson.JsonParser.parse(JsonParser.java:98)
	at au.com.dius.pact.provider.ResponseComparison$Companion.generateFullDiff(ResponseComparison.kt:91)
	at au.com.dius.pact.provider.ResponseComparison$Companion.access$generateFullDiff(ResponseComparison.kt:85)
	at au.com.dius.pact.provider.ResponseComparison.bodyResult(ResponseComparison.kt:79)
	at au.com.dius.pact.provider.ResponseComparison$Companion.compareResponse(ResponseComparison.kt:124)
	at au.com.dius.pact.provider.ProviderVerifier.verifyRequestResponsePact(ProviderVerifier.kt:476)
	at au.com.dius.pact.provider.junit5.PactVerificationContext.validateTestExecution(PactJUnit5VerificationProvider.kt:96)
	at au.com.dius.pact.provider.junit5.PactVerificationContext.verifyInteraction(PactJUnit5VerificationProvider.kt:77)
```

https://github.com/DiUS/pact-jvm/blob/master/provider/pact-jvm-provider/src/main/kotlin/au/com/dius/pact/provider/ProviderClient.kt#L363-L367

PabloThiele:
Adding some support on the discussion: [HTTP RFC](https://www.w3.org/Protocols/rfc2616/rfc2616-sec7.html#sec7.2.1)
[...]

PR #1046, opened 2020-03-12
#1013 - Fix to issue when in pact contract content type is not provided

wieslawmlynarski:
When the body is of type JSON use the application/json content type
Breaking change between pact 4.0.6 -> 4.0.7


uglyog:
Thanks for the PR!

PR #1048, opened 2020-03-13
feat: create Kotlin friendly extension for arrays

gmariotti:
This PR adds a Kotlin extension to make more friendly adding an array inside an object.

For example the following code:
```kotlin
newObject {
  array("array") {
    it.stringType("property")
  }
}
```
can now be written:
```kotlin
newObject {
  newArray("array") {
    stringType("property")
  }
}
```

Also, I've fixed the dependencies for the `pact-jvm-consumer-java8` module. The Junit 5 tests were not running with the previous setup.


uglyog:
Thanks for the PR!

Issue #1049, opened 2020-03-16
Pact JUnit provider not publishing verification results if Pact URL is overridden


arhohuttunen:
I added a workflow where a consumer triggers the verification on the provider side, and overrides the Pact URL with `pact.filter.pacturl`. The verification is being run correctly, but publishing the 
results is being skipped.

Running the Gradle task with info level:

```
[Test worker] INFO  a.c.d.p.p.DefaultVerificationReporter.reportResults - Skipping publishing verification results for source UrlSource
```

I did some digging and found out this in `DefaultVerificationReporter.kt`:

```kotlin
  override fun reportResults(
    pact: Pact<out Interaction>,
    result: TestResult,
    version: String,
    client: PactBrokerClient?,
    tag: String?
  ) {
    when (val source = pact.source) {
      is BrokerUrlSource -> {
        val brokerClient = client ?: PactBrokerClient(source.pactBrokerUrl, source.options)
        publishResult(brokerClient, source, result, version, pact, tag)
      }
      else -> logger.info { "Skipping publishing verification results for source $source" }
    }
  }
```

So, if the type of the Pact source is not `BrokerUrlSource`, it skips the publishing of the results.

I think the culprit is most likely this piece of code in `PactBrokerLoader.kt`:

```kotlin
  override fun load(providerName: String): List<Pact<Interaction>> {
    val resolver = setupValueResolver()
    val pacts = when {
      overriddenPactUrl.isNotEmpty() -> {
        val pactBrokerClient = newPactBrokerClient(brokerUrl(resolver).build(), resolver)
        val pactSource = UrlSource<Interaction>(overriddenPactUrl!!)
  // ...
```

So, basically when we override the Pact URL, the Pact source here is created as an `UrlSource` which results into the verification results being skipped.

uglyog:
4.0.8 released with this change

PR #1050, opened 2020-03-16
fix: publish verification results when Pact URL is overridden #1049

arhohuttunen:
Fixes the problem where `@AllowOverridePactUrl` and `pact.filter.pacturl` are used together with `pact.verifier.publishResults=true` but the results are not published.


uglyog:
Thanks for the PR!

Issue #1023, opened 2020-02-21
@PactBroker not reading Spring properties with JUnit 5


arhohuttunen:
If you provide system properties in Spring `application.yml` these won't be read by the `@PactBroker` annotation when running with JUnit 5. This works fine if you provide them as normal system properties.

This issue seems to have been fixed for the JUnit 4 runners in here: https://github.com/DiUS/pact-jvm/issues/632

I can see that the problem has been fixed by adding a property resolver, and that it is also possible to pass a `valueResolver` attribute for the `@PactBroker` annotation. However, since you need a Spring context to be able to create such a resolver for Spring, it cannot be done with the annotation.

I also tried to add a resolver into `PactVerificationContext` in `@BeforeEach` but that didn't seem to do the trick.

vtapadia:
I am also having the same issue.
The preferred approach to ask questions on stackflow did not help either.
https://stackoverflow.com/questions/60153108/provider-test-integration-with-pact-broker-for-spring-boot-junit5-configuratio

Waiting for the answer

uglyog:
I've created a new module pact-jvm-provider-junit5-spring with a TestTemplateInvocationContextProvider that extends the current JUnit 5 one (PactVerificationSpringProvider). This should allow you to configure all the properties in the Spring context.

arhohuttunen:
That’s awesome! Thank you for such a quick response to the issue. This will make the developer experience smoother when running tests out of the box, with the broker having a default configuration in the properties file.

vtapadia:
HI @uglyog , I noticed a small inconsistency in the implementation of the [SpringEnvironmentResolver.kt ](https://github.com/DiUS/pact-jvm/blob/master/provider/pact-jvm-provider-junit5-spring/src/main/kotlin/au/com/dius/pact/provider/spring/junit5/SpringEnvironmentResolver.kt)

The `propertyDefined()` method relies only on environment, while the `resolveValue() ` uses SystemPropertyResolver also as backup. 
Also, the Spring _environment_ property resolution considers the system properties also when resolving a value. That has a higher preference. Any specific reason why?

uglyog:
@vtapadia I've removed the SystemPropertyResolver fallback

uglyog:
@vtapadia removing the fallback breaks the pact loader, so I'm reverting the change

arhohuttunen:
[...]

PR #1051, opened 2020-03-16
fix: @PactBroker not reading Spring properties with JUnit 5 #1023

arhohuttunen:
Fixes an issue in `PactVerificationInvocationContextProvider.resolvePactSources()` where calling `PactBrokerLoader.description()` ignores the value resolver that is being set after the call.

The method calling order breaks the `PactVerificationSpringProvider` that is trying to override the `getValueResolver()` method to be able to load broker properties from a Spring Boot `application.yml` file.

### DIFF
--- a/provider/pact-jvm-provider-junit5/src/main/kotlin/au/com/dius/pact/provider/junit5/PactJUnit5VerificationProvider.kt
+++ b/provider/pact-jvm-provider-junit5/src/main/kotlin/au/com/dius/pact/provider/junit5/PactJUnit5VerificationProvider.kt
@@ -392,11 +392,11 @@ open class PactVerificationInvocationContextProvider : TestTemplateInvocationCon
     logger.debug { "Verifying pacts for provider '$serviceName' and consumer '$consumerName'" }
 
     val pactSources = findPactSources(context).flatMap {
-      description += "\nSource: ${it.description()}"
       val valueResolver = getValueResolver(context)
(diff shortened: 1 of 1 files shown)


arhohuttunen:
I was also thinking about some kind of test for it, but could not easily figure out a clean way to do it.

What do you think about adding `spring-boot-starter-test` and `spring-boot-starter-web` as test dependencies into the `pact-jvm-provider-junit5-spring` module?

Then you could have a kind of an integration test with this kind of approach:

`src/test/resources/application.yml`:
```yml
pactbroker:
  host: localhost
  port: ${local.server.port}
```

`PactVerificationSpringProviderTest.kt`:
```kotlin
@ExtendWith(SpringExtension::class)
@SpringBootTest(webEnvironment = SpringBootTest.WebEnvironment.RANDOM_PORT)
@Provider("Animal Profile Service")
@PactBroker
@IgnoreNoPactsToVerify
internal class PactVerificationSpringProviderTest {
    @TestTemplate
    @ExtendWith(PactVerificationSpringProvider::class)
    fun pactVerificationTestTemplate(context: PactVerificationContext?) {
        context?.verifyInteraction()
    }
}
```

uglyog:
Thanks for the PR!

You're test method is clever, and I can confirm that it works. I have added those changes to the module.

uglyog:
4.0.8 released with this change

PR #1071, opened 2020-04-15
feat: add namespace-aware XML matching

asteffey:
## Problem to solve

XML namespaces are not currently used when comparing nodes between XML documents.  This poses a problem when different services assign different prefixes for the same namespace. For example, if `<a:foo xmlns:a="urn:ns"/>` is expected, then `<foo xmlns="urn:ns"/>` or `<b:foo xmlns:b="urn:ns"/>` will not currently match.

## Proposed changes

XML namespaces will be utilized when comparing nodes between XML documents.

When an XML namespace is defined (e.g., `<prefix:local xmlns:prefix="urn:ns"/>`), 
nodes will be compared by their namespace (e.g., `urn:ns`) and the local part of their name (e.g., `local`). Nodes using different
XML namespace declarations/prefixes (e.g., `<foo xmlns="urn:ns"/>`, `<a:foo xmlns:a="urn:ns"/>`, `<b:foo xmlns:b="urn:ns"/>`) can match.

When XML namespaces are not defined (e.g., `<prefix:local/>`), nodes will be compared by their full node name which includes both the local part
and optional prefix (e.g., `prefix:local`). Un-namespaced nodes with different full node names (e.g., `<foo/>`, `<a:foo/>`, `<b:foo/>`) won't match.

Nodes without an XML namespace (e.g., `<foo/>`) should not match nodes with an XML namespace (e.g., `<foo xmlns="urn:ns"/>`).

Namespace-aware matching can be disabled by setting the optional `pact.matching.xml.namespace-aware` system property to false (it will be considered true by default).

## NOTICE

### DIFF
--- a/core/matchers/src/main/kotlin/au/com/dius/pact/core/matchers/MatcherExecutor.kt
+++ b/core/matchers/src/main/kotlin/au/com/dius/pact/core/matchers/MatcherExecutor.kt
@@ -34,7 +34,7 @@ fun valueOf(value: Any?): String {
   return when (value) {
     null -> "null"
     is String -> "'$value'"
-    is Element -> "<${value.tagName}>"
+    is Element -> "<${QualifiedName(value)}>"
     is Text -> "'${value.wholeText}'"
     else -> value.toString()
   }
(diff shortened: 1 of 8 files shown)


asteffey:
Because this is a change to current behavior, in addition to having the system property to disable it, I also have some proposed text to update the Pact Specification, currently targeting v4: https://github.com/mitre/pact-specification/commit/5daf7c7412758ebb6b9c42fd8a74a7ff13502e90

I can create a second PR for this as well.

uglyog:
That's awesome, I'll port this change to the Rust code base.

uglyog:
We can add those specification tests to the V3 version, because it is actually correcting the behaviour.

uglyog:
@mefellows we need to port this to the Rust implementation

Issue #1078, opened 2020-04-27
Provider Tests will fail when optional generator attributes are missing


ankaubisch:
With version 4.0.10 there is a bug that provider tests will fail, when a pact contract contains a date or time related generator without the optional attribute `format`.

This is caused by the current implementation of `Json.toString` that will return a "null" String when `JsonElement` argument is null. 
`DateGenerator`, `DateTimeGenerator` and `TimeGenerator` will expect a null value or a correct pattern, but will receive a "null" `String ` which fails with following message:

> line 1:0 token recognition error at: 'nu'
> line 1:2 token recognition error at: 'll'
> line 1:4 mismatched input '<EOF>' expecting {'next', 'last', 'now', 'today', 'yesterday', 'tomorrow', '+', '-'}

PR #1079, opened 2020-04-27
fixes an issue where all date and time based generators will crash provider tests

ankaubisch:
This bugfix will introduce a new function in `Json` object  called `toNullableString`that will return null when `JsonElement` is null.

This function will be used in `DateGenerator`, `TimeGenerator` and `DateTimeGenerator`.

Tests are included to show the behavior of parsing json without optional "format" arguments.


ankaubisch:
#1078

uglyog:
Thanks for the PR!

PR #1095, opened 2020-05-20
Add SLF4JReporter

Koriit:
Hello,
In my company we are using automatic pact verification with [report portal](https://reportportal.io/) for easy time checking what went wrong for both consumers and providers.

Unfortunately, provider output does not log and only logs are transferred to report portal.

Thus, I had to implement my custom reporter that outputs to SLF4J logger. Then, I thought it would be nice if such reporter was available out of the box as right now we have to move it between different projects. Hence the pull request. This version is adjusted to the latest changes on the master.

The reporter should be roughly equivalent to the console one. I hope I didn't miss anything and it can be integrated smoothly!


Koriit:
Ok, didn't realize there were some breaking changes on the master. :D Gonna fix it first.

Koriit:
Ok, fixed. Though, master seems to be broken and tests don't finish now.

uglyog:
Awesome, thanks for the PR!

I released 4.1.0 to master yesterday, which has all the modules renamed. I think you might have been caught by that.

PR #1109, opened 2020-06-02
add helper function for MessagePacts

chrisport:
the helper function return the verification result and write the pact to a file

Looking forward to any comments.


uglyog:
LGTM

Issue #1110, opened 2020-06-02
Encoding error when contract generating for request with binary body (for images)


weofferservice:
When request has binary body (for images) contract is generating to JSON in UTF-8 encoding.
This is error, because UTF-8 irreversible.
You must use Base64 or ISO_8859_1
I found this bug and fix it.
I will make pull request to patch this error and attach link here.

I attached to this Issue my test data which helped me to find bug:
1) ![face-with-smile.jpg](https://user-images.githubusercontent.com/37179306/83542947-57d12d00-a504-11ea-8df4-de601c9109c1.jpg)
2) PhotoQualityOkEndpointTest.java
`package com.example;

import au.com.dius.pact.consumer.MockServer;
import au.com.dius.pact.consumer.dsl.PactDslJsonBody;
import au.com.dius.pact.consumer.dsl.PactDslWithProvider;
import au.com.dius.pact.consumer.junit5.PactConsumerTestExt;
import au.com.dius.pact.consumer.junit5.PactTestFor;
import au.com.dius.pact.core.model.RequestResponsePact;
import au.com.dius.pact.core.model.annotations.Pact;
import org.apache.commons.lang3.StringUtils;
import org.assertj.core.api.Assertions;
import org.json.JSONArray;
import org.json.JSONException;
import org.json.JSONObject;
import org.junit.jupiter.api.Test;
import org.junit.jupiter.api.extension.ExtendWith;
import org.skyscreamer.jsonassert.JSONAssert;
import org.springframework.beans.factory.annotation.Value;
import org.springframework.boot.test.context.ConfigFileApplicationContextInitializer;
import org.springframework.core.io.ByteArrayResource;
import org.springframework.http.*;
import org.springframework.test.context.ContextConfiguration;
import org.springframework.test.context.junit.jupiter.SpringExtension;
import org.springframework.test.util.ReflectionTestUtils;
import org.springframework.util.LinkedMultiValueMap;
import org.springframework.util.MultiValueMap;
import org.springframework.web.client.RestTemplate;

import java.io.IOException;
import java.nio.charset.StandardCharsets;
import java.util.Collections;
import java.util.Map;

import static com.example.PactFisengineConstants.*;

@ExtendWith({PactConsumerTestExt.class, SpringExtension.class})
@ContextConfiguration(initializers = ConfigFileApplicationContextInitializer.class)
@PactTestFor(providerName = PROVIDER, hostInterface = "localhost", port = "8080")
[...]

PR #1111, opened 2020-06-02
Issue-1110: fixing "Encoding error when contract generating for reque…

weofferservice:
fixing https://github.com/DiUS/pact-jvm/issues/1110


uglyog:
Looks like `ByteArrayOutputStream.toString(Charset)` is not available in JDK 8 and 9. That is why the Travis build has failed.

We don't need to go via a String, we could just put the byte array straight into the OptionalBody.

uglyog:
I'm closing this PR as issue #1110 has been addressed with commits f67ecbbee3449e930a6dcc9d9c262d286fe94206 and dde272b05d419f190de3d2c170425eee79b97ed9

PR #1123, opened 2020-06-14
fix: correct display name for junit5 verification provider

anto-ac:


anto-ac:
@uglyog The displayName was only showing the consumerName.

I'm not sure if this change will be correct under all circumstances. I've tested it locally for our case of tagged, non-pending pacts coming form the broker.

My use of kotlin might be dodgy here... let me know if you think that the implementation makes sense and if it could be re-written more elegantly.  

I haven't found any tests for the displayName. Should we add some?

provider/junit5/src/main/kotlin/au/com/dius/pact/provider/junit5/PactJUnit5VerificationProvider.kt
```diff
@@ -155,12 +155,15 @@ class PactVerificationExtension(
...
+      pactSource is BrokerUrlSource && pactSource.result != null -> {
+        var displayName = pactSource.result!!.name + " - ${interaction.description}"
+        if (pactSource.tag.isNotEmpty()) displayName += " (tag ${pactSource.tag})\""
+        if (pactSource.result!!.pending) {
+          "$displayName [PENDING]"
+        } else {
+          displayName
+        }
       }
-      pactSource is BrokerUrlSource && pactSource.tag.isNotEmpty() -> "${pact.consumer.name} - ${interaction.description} (tag ${pactSource.tag})"
```
uglyog, line 163: We still need this case, because the pactSource.result value is only populated from the new broker endpoint. For people with an older broker, it will be null.

anto-ac: Added the case back!

PR #1127, opened 2020-06-15
fix: backward compatible support for Z UTC designator in timestamps

anto-ac:
@uglyog what do you think about this? Perhaps a bit hacky, but at the same time there might be other organisations in a similar situation to ours.

### DIFF
--- a/core/matchers/src/main/kotlin/au/com/dius/pact/core/matchers/MatcherExecutor.kt
+++ b/core/matchers/src/main/kotlin/au/com/dius/pact/core/matchers/MatcherExecutor.kt
@@ -324,7 +324,11 @@ fun <M : Mismatch> matchDateTime(
     emptyList()
   } else {
     try {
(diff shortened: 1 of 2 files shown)


uglyog:
Although it feels like a bad thing to do, it looks like it might be ok. It looks like ISO 8601 Time zone designation is a superset of the RFC 822 one. But maybe we should output a warning that the substitution has been made?

anto-ac:
Good point on the warning. 

I’ve just realised that my change wouldn’t take into account formats with quotes though, and I think the parsing would fail in that case.

uglyog:
Ah, yeah, that's a problem.

I'm thinking we can add a JVM system property that turns on the behaviour. That way it is opt in.

anto-ac:
Or... I was thinking... a regex that substitutes only Z that are not enclosed in quotes

anto-ac:
@aplsup @e-ivaldi @uglyog If I could get your reviews/approvals, that would be great.

PR #1133, opened 2020-06-17
Issue with object with decimal value

anto-ac:
See new tests added in `JsonParserSpec`. Most decimal values as values were not being parses correctly. 

`{"1": { "2": 200.25}}`: was working
`{"1": { "2": 20.25}}`: wasn't working.

The method changed by @e-ivaldi is used in a lot of places, so @uglyog could you think of any unwanted effects?


core/support/src/main/java/au/com/dius/pact/core/support/json/BaseJsonLexer.java
```diff
@@ -94,7 +94,7 @@ protected void skipWhitespace() {
   }
 
   private char[] allocate(char[] buffer, int size) {
-    char[] newBuffer = new char[Math.max(buffer.length * 2, size)];
```
uglyog, line 97: The actual change I was trying to do should be:
`buffer.length + Math.max(buffer.length, size)`

e-ivaldi: Ah thank you, what's the rationale behind that?
I have mostly looked at it within the context of parsing the decimal numbers, where as long as the buffer is of the proper size everything's fine, and only checked all the tests were still good with it.

uglyog: I've been trying to minimise the number of array allocations and copies for performance reasons. So start with 16 characters, and then double it if we need to extend. That way it should probably only happen once for numbers. But for large strings it will go from the base (128), 256, 512, 1024, etc.

uglyog: For that case, we have parsed the first part of the number, so size is that number of digits, but there will be more after the `.`, so just expand the buffer to the next point as long as that is big enough.

e-ivaldi: Thank you for expanding on it

PR #1135, opened 2020-06-18
maven-plugin: adds support to configure pact-publish via command line…

ivangsa:
we would like to use maven plugin to publish pacts to pact broker without adding the plugin to the pom..


uglyog:
Does this require everyone to use the JVM properties, or will it still work with configuring it in the POM?

ivangsa:
Pom configuration will still work as expected..
JVM properties will work in addition to pom configuration... if they are not configured in pom they can be passed from the command line

uglyog:
Awesome!

PR #1137, opened 2020-06-18
fix: content-type super-type resolution for application/octet-stream …

anto-ac:
…is now correct

### DIFF
--- a/core/model/src/main/kotlin/au/com/dius/pact/core/model/ContentType.kt
+++ b/core/model/src/main/kotlin/au/com/dius/pact/core/model/ContentType.kt
@@ -48,7 +48,7 @@ class ContentType(val contentType: MediaType?) {
 
   fun isBinaryType(): Boolean {
     return if (contentType != null) {
-      val superType = registry.getSupertype(contentType)
+      val superType = registry.getSupertype(contentType) ?: MediaType.OCTET_STREAM
       val type = contentType.type
       val baseType = superType.type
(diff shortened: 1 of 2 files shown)


core/model/src/main/kotlin/au/com/dius/pact/core/model/ContentType.kt
```diff
@@ -48,7 +48,7 @@ class ContentType(val contentType: MediaType?) {
 
   fun isBinaryType(): Boolean {
     return if (contentType != null) {
-      val superType = registry.getSupertype(contentType)
+      val superType = registry.getSupertype(contentType) ?: MediaType.OCTET_STREAM
```
e-ivaldi, line 51: `    /**
     * Returns the supertype of the given type. If the media type database
     * has an explicit inheritance rule for the type, then that is used. 
     * Next, if the given type has any parameters, then the respective base 
     * type (parameter-less) is returned. Otherwise built-in heuristics like 
     * text/... -&gt; text/plain and .../...+xml -&gt; application/xml are used. 
     * Finally application/octet-stream is returned for all types for which no other
     * supertype is known, and the return value for application/octet-stream
     * is <code>null</code>.
     *
     * @since Apache Tika 0.8
     * @param type media type
     * @return supertype, or <code>null</code> for application/octet-stream
     */
    public MediaType getSupertype(MediaType type) {`

aplsup:
LGTM :+1:

Issue #1130, opened 2020-06-16
Pact Broker on a path


jgelon:
I am trying to get my pact broker working on my environment. I have the broker running in K8S under https://mydomain/pactbroker.
I am able to send to the broker with the maven plugin (publish). However when I try to verify I get an error; Request to path '/' failed with response 'HTTP/1.1 401 Unauthorized'

Can someone help me out?

```
   <build>
        <plugins>
            <plugin>
                <groupId>au.com.dius</groupId>
                <artifactId>pact-jvm-provider-maven</artifactId>
                <version>4.0.10</version>
                <configuration>
                    <serviceProviders>
                        <!-- You can define as many as you need, but each must have a unique name -->
                        <serviceProvider>
                            <name>FaqService</name>
                            <protocol>http</protocol>
                            <host>localhost</host>
                            <port>8080</port>

                            <pactBroker>
                                <url>https://mydomain/pactbroker/</url>
                                <authentication>
                                    <scheme>basic</scheme>
                                    <username>user</username>
                                    <password>pass</password>
                                </authentication>
                            </pactBroker>
                        </serviceProvider>
                    </serviceProviders>
                </configuration>
            </plugin>
        </plugins>
    </build>

```

uglyog:
`HTTP/1.1 401 Unauthorized` means your broker rejected the request because the authentication was wrong. You need to check the username and password, as well as check you can access your server with something like curl.

uglyog:
Also, some debug logs might help as well.

jgelon:
When trying to go through the logs it seems it tries to fetch the HAL root information via path "/". However responds with; [WARNING] Could not fetch the root HAL document

When I enable PreEmptive Authentication i can see that ot give a Warning like [WARNING] Using preemptive basic authentication with the pact broker at https://mydomain so without the path.

jgelon:
I also tried with and without credentials, in both cases same issue

jgelon:
Have been looking a bit more in debug mode and found the issue being the encodedPath;

au.com.dius.pact.core.pactbroker.util.HttpClientUtils
```
      if (encodePath) {
        URIBuilder(baseUrl).setPath(url).build()
      } else {
        URI(baseUrl + url)
      }
```

It is going in as baseUrl: http://mydomain/pactbroker and url: / with encodePath=true which results in my case in http://mydomain/
Can I disable encodePath because it will work then

jgelon:
[...]

PR #1139, opened 2020-06-18
Fix: Keep existing path on baseUrl

jgelon:
Keep path on baseUrl when combining baseUrl and url instead of throwing it away.
Solves issue #1130


uglyog:
Thanks for the PR!

jgelon:
@uglyog I see this has been merged. However it was not part of the release 4.1.3....

uglyog:
I've just released 4.1.5

PR #1152, opened 2020-07-01
chore: improve logging when falling back to DateUtils.parseDate

anto-ac:


core/matchers/src/main/kotlin/au/com/dius/pact/core/matchers/MatcherExecutor.kt
```diff
@@ -354,9 +354,10 @@ fun <M : Mismatch> matchDateTime(
     } catch (e: DateTimeParseException) {
       try {
         logger.warn {
-          """Unable to parse ${valueOf(actual)} with $pattern using java.time.format.DateTimeFormatter.
-          Attempting to parse using org.apache.commons.lang3.time.DateUtils 
-          to guarantee backwards compatibility with versions < 4.1.1.
+          """Failed to parse ${valueOf(actual)} with '$pattern' using java.time.format.DateTimeFormatter.
```
uglyog, line 357: You can pass the exception to the logger method as well

```
logger.warn(e) {
```

PR #1156, opened 2020-07-02
fix: fix a problem where the provider version trim property is ignored

e-ivaldi:


anto-ac:
Given the last few instances, @e-ivaldi and I would like to test this locally before merging, but we can't because `master` is broken.

provider/src/main/kotlin/au/com/dius/pact/provider/ProviderVersion.kt
```diff
@@ -0,0 +1,37 @@
...
+
+  override fun get(): String {
+    val version = source() ?: FALLBACK_VALUE
+
+    if (version == FALLBACK_VALUE) {
+      logger.warn("Provider version not set, defaulting to '$FALLBACK_VALUE'")
+    }
+
+    val trimSnapshotProperty = System.getProperty(ProviderVerifier.PACT_PROVIDER_VERSION_TRIM_SNAPSHOT)
+    val isTrimSnapshot: Boolean = if (trimSnapshotProperty.isNullOrBlank()) false else trimSnapshotProperty.toBoolean()
```
e-ivaldi, line 27: type can be removed

anto-ac:
Tested with our services and it works

PR #1158, opened 2020-07-02
fix: Omitting consumer version selectors should fall back to tags

### LINKED ISSUE #1157: Cannot run provider verification tests when using @PactBroker(tags = ...) (since 4.1.5)
No pacts are found. 

Introduced by one of these PRs:

https://github.com/DiUS/pact-jvm/pull/1147
https://github.com/DiUS/pact-jvm/pull/1143

@pendsley

pendsley:
Fixes #1157 

Tested locally with:
No tags, No consumer version selectors
Only tags, no version selectors
No tags, Only version selectors
Tags and version selectors (Should prefer version selectors)

### DIFF
--- a/provider/junit/src/test/groovy/au/com/dius/pact/provider/junit/loader/PactBrokerLoaderSpec.groovy
+++ b/provider/junit/src/test/groovy/au/com/dius/pact/provider/junit/loader/PactBrokerLoaderSpec.groovy
@@ -720,16 +720,13 @@ class PactBrokerLoaderSpec extends Specification {
       loader.pactReader = mockReader
       loader
     }
-    def selectors = [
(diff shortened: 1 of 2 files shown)


provider/junit/src/test/groovy/au/com/dius/pact/provider/junit/loader/PactBrokerLoaderSpec.groovy
```diff
@@ -653,16 +720,13 @@ class PactBrokerLoaderSpec extends Specification {
       loader.pactReader = mockReader
       loader
     }
-    def selectors = [
-      new ConsumerVersionSelector('latest', true)
```
pendsley, line 657: In hind sight, this should've been a red flag

PR #1165, opened 2020-07-08
Add note about not being able to publishing results to @PactFilter se…

tinexw:
…ction

### DIFF
--- a/provider/junit/README.md
+++ b/provider/junit/README.md
@@ -393,6 +393,8 @@ public class PactJUnitTest {
 }
 ```
 
+**NOTE!** You will only be able to publish the verification results if all interactions have been verified. If an interaction is not covered because it was filtered out, you will not be able to publish.
+
 ##### Filtering the interactions that are run
 
 **(version 4.1.2+)**


provider/junit/README.md
```diff
@@ -393,6 +393,8 @@ public class PactJUnitTest {
 }
 ```
 
+**NOTE!** You will not be able to publish the verification results when you apply a filter.
```
e-ivaldi, line 396: Is it worth specifying that it only happens if you leave one or more interaction outside of your filters?

Potentially you can have multiple filters that cover all the interactions and in that case the verification results will be published, right?

tinexw: Hmmm.. I realized I tested this in the 3.5.x branch (because I was investigating another bug). It might no longer be the case in 4, I will check.

tinexw:
Closing until I checked the current version.

tinexw:
The behavior actually changed in 4. It is now possible to publish when not all interactions were verified.

anto-ac:
Which should not happen. I think there is a bug that I haven’t been able to create a reproduceable example for.

A verification result should be published only if all interactions have been verified (successfully or not)

tinexw:
Well.. strange. When using maven publishing is not allowed when using a filter.

tinexw:
> Which should not happen. I think there is a bug that I haven’t been able to create a reproduceable example for.
> 
> A verification result should be published only if all interactions have been verified (successfully or not)

Yes, there seems to be a bug. I'm currently checking junit4 in the 4.x version. 
I have one branch with 3.5.x and junit4 where publishing never works if filters are applied, even if combined they cover all interactions.
And then I have another branch with 4.x and junit5 were publishing always works. I will check out the 4.x / junit4 combination next.
Also I will check for maven and the combined filters.

tinexw:
So I investigated more. I only looked at the most recent Pact version, no longer 3.5.x. It is broken for Junit5. For Junit4 and Maven it works as expected. Should I create a bug?

Artifact| One interaction filtered out, other missing| all interactions covered 
 ----------- | -----------  |  ----------- 
JUnit4 | not published :white_check_mark: | published :white_check_mark:
JUnit5 | published :x: |  published :white_check_mark:
Maven | not published :white_check_mark: | published :white_check_mark:

anto-ac:
If interactions are filtered and not all interactions have been verified it should NOT publish verification results.

provider/junit/README.md
tinexw: I extended the description accordingly.

tinexw:
Since the behavior is now clear, I updated the README accordingly and reopened this PR.

provider/junit/README.md
e-ivaldi: Brilliant, thanks for looking into that, are you going to open an issue about the publication bug with junit5 ?

e-ivaldi: I will take a look to see why that's happening (if you are not already planning to do so yourself)

tinexw: Yes, looking into fixing it right now, will either create issue + PR or just the issue if I cannot make any progress.

e-ivaldi: Okay :+1:  :+1:  :+1:

uglyog:
Thanks for the PR. Not that anyone will read the docs ....

Issue #1166, opened 2020-07-09
Verification results are published even though not all interactions were verified


tinexw:
When using JUnit 5 in combination with `@FilteredPact`, the verification results are **always** published even if not all interactions were verified. See also discussion in [this PR](https://github.com/DiUS/pact-jvm/pull/1165#issuecomment-656290908).

Right now to me it looks like this is just not implemented for JUnit 5. I will look into it.

uglyog:
4.1.7 has been released

PR #1167, opened 2020-07-09
fix: Don't publish incomplete verification results #1166

tinexw:
Fixes https://github.com/DiUS/pact-jvm/issues/1166

What is the best way to test this? I would like to add tests similar to [the JUnit 4 ones](https://github.com/DiUS/pact-jvm/blob/master/provider/junit/src/test/groovy/au/com/dius/pact/provider/junit/InteractionRunnerSpec.groovy#L102). But I don't know how.


anto-ac:
I’m not familiar with that area of the code to comment on the implementation, but it would be good to see some tests added, at least to reproduce/cover the fix.

tinexw:
> I’m not familiar with that area of the code to comment on the implementation, but it would be good to see some tests added, at least to reproduce/cover the fix.

I can add the most basic test to ensure that respective methods are called. I just thought it would be better to also add a test on an integration test level.

tinexw:
I added the simple test.

PR #1182, opened 2020-08-01
Feature/webflux support

paweusz:
Adds support for testing reactive spring components (webflux) on server side. 
Uses `WebTestClient` for testing routers and controllers.
Counterpart of MockMvc testing support.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -41,7 +41,7 @@ subprojects {
     return
   }
 
-  version = '4.1.8-SNAPSHOT'
+  version = '4.1.7'
   
   buildscript {
     repositories {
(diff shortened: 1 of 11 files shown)


build.gradle
```diff
@@ -41,7 +41,7 @@ subprojects {
     return
   }
 
-  version = '4.1.7'
```
uglyog, line 44: Please don't add snapshot to the version

uglyog: 4.1.7 will be the next version to be released

paweusz: Thx, changed back to 4.1.7

PR #1184, opened 2020-08-03
Ignore order json

jstandard:
## Problem to solve

Currently, the list items of an actual response must be in the same order as list items in the expected response.
This is a problem when a consumer expects a specific set (or subset) of values in a list, but does not
care about order of elements (or potentially any extra values in the set).

## Proposed changes

- Adds an EqualsIgnoreOrderMatcher which will uniquely match each item of an actual array against the
  expected array, regardless of the order of the actual array items.
  - For example, `[2,3,1]` would match `[1,2,3]`.
  - The EqualsIgnoreOrderMatcher defaults to Equality matching for non-arrays, such as the items of
    the array.
  - Matching rules can be defined for child items of the array using array indexes in
    the path (corresponding to the expected item indexes), or for all child items of the array
    using a wildcard in the path.
  - EqualsIgnoreOrderMatcher expects the actual array to be the same size as the expected array.

- Adds Min/Max/MinMaxEqualsIgnoreOrderMatcher for cases where array sizes can differ.
  - Any extra items in the array will only be matched if a wildcard matcher is defined for the array
    items.

- When ignore-order array matching is being performed, a '*' is used instead of numeric index 
  values when reporting wildcard mismatches.
  - This is intended to show the cause of the mismatch more clearly, but is inconsistent with other
    cases of reporting wildcard mismatches.
  - Because ignore-order matching must consider all permutations of the actual values, the numeric
    index value wouldn't be meaningful.
  - As this functionality is a little inconsistent with the existing mismatch reporting behavior,
    it was separated into its own commit so it could be easily dropped/reverted.

- Fixed some small detekt issues in core:matchers
  - we triggered a TooManyFunctions warning in `Matchers` and ComplexMethod in
    `MatcherExecutor.domatch` and there wasn't a clear way to correct these
    problems without a significant refactor of code outside of our changes.

## NOTICE


consumer/src/test/groovy/au/com/dius/pact/consumer/dsl/DslPartSpec.groovy
```diff
@@ -166,7 +206,8 @@ class DslPartSpec extends Specification {
     'matchMin'        | 1            | [match: 'type', min: 1]
     'matchMax'        | 1            | [match: 'type', max: 1]
     'includesMatcher' | 1            | [match: 'include', value: '1']
-
+    'matchMinIgnoreOrder' | 1        | [match: 'ignore-order', min: 1]
```
uglyog, line 209: I think this should be `equals-ignore-order` to be more explicit

uglyog:
This looks useful!

I've added an item to the Pact V4 RFC. The main problem with merging this, pacts using this new matcher will not be compatible with the other language implements. Also, probably better to create a 4.2.x branch for this to be merged into to

consumer/src/test/groovy/au/com/dius/pact/consumer/dsl/DslPartSpec.groovy
uglyog: Actually, ignore that. It's not only equals, but also other matchers.

uglyog:
I have merged this into the v4.2.x branch. I'll also add a check in the mock server to verify that the current pact spec version is V4 when this feature is used.

PR #1199, opened 2020-08-26
Feature/publish to broker

markozz:
Hi,

With this pull request I would like to contribute the following feature to the pact-JVM-server;
* publish a contract which is created on disk to a pact broker.

By running the pact-jvm-server with parameter -b "https://valid.pact.broker.com" and optionally -t "val!dT0ken" a call to the endpoint `/publish` will push the contract from disk to the broker.

Please let me know what you think!

Kind regards,

Mark


uglyog:
LGTM.

There is a [PactBrokerClient](https://github.com/DiUS/pact-jvm/blob/master/core/pactbroker/src/main/kotlin/au/com/dius/pact/core/pactbroker/PactBrokerClient.kt#L210) class in core/pactbroker module that would be better to use, as it uses HAL links navigation instead of templating the URI and also supports tagging the pact version.

Here is an [example](https://github.com/DiUS/pact-jvm/blob/master/provider/gradle/src/main/groovy/au/com/dius/pact/provider/gradle/PactPublishTask.groovy#L47) of it being used.

I guess the tags would be an extra attribute on the body.

markozz:
Hi, implemented your suggestions.

uglyog:
Awesome, thanks!

Issue #1193, opened 2020-08-24
Filtering by Consumers when loading from PactBroker is broken


marcos-scholtz:
When loading pacts from the PactBroker, one can filter them by consumer(s) using the system-property:
```
-Dpactbroker.consumers=my-consumer-name
```

When doing this however, it always results in a "NoPactsFoundException". Without this parameter, it does find the pact correctly for this consumer.

The reason seems to be in the PactBrokerLoader class, in line 195:
```
      if (pactBrokerConsumers.isNotEmpty()) {
        val consumerInclusions = pactBrokerConsumers.flatMap { parseListExpression(it, resolver) }
        consumers = consumers.filter { consumerInclusions.isEmpty() || consumerInclusions.contains(it.name) }
      }
```

According to my tests, "consumers" is a List of "PactBrokerResult" instances. The PactBrokerResult has a "name" field, but it contains the following text:
```
Pact between my-consumer-name (d49036c9b747af48c33fc66de0ddf7af57756c49) and my-provider-name
```
This text does NOT match the "consumerInclusions", they merely contain "my-consumer-name". 

One solution could be to simply check that "my-consumer-name" is contained in the result's name field in any position, but this sounds hacky, it would probably be better to load the specific pact from the broker and check the real consumer name, but that's one call more for each consumer.

artemptushkin:
Same for me

`au.com.dius.pact.provider:junit:4.1.7`

artemptushkin:
@uglyog please take a look at the created PR for this

anto-ac:
I think I've hit exactly the same problem when I upgraded the pact broker earlier today, and it's down to which api is used to retrieve the pacts, with the old api removed when pact for verifications came out of beta.

@artemptushkin and @marcos-scholtz - just to confirm, with which version of the broker you run into this?

artemptushkin:
@anto-ac I think the latest `docker pull dius/pact-broker:2.63.0.0`

uglyog:
I've updated the pact broker loader to send the consumer names in the selectors if using the new endpoint. The pact broker should do the filtering. When using the old end-point, it works as before.

anto-ac:
Has anyone had a chance to verify the changes? @artemptushkin maybe? If they work as expected, I'd be happy to release.

artemptushkin:
@anto-ac I could, but whats the pact-broker version to test @uglyog ? It doesn't depend on any `pact-jvm` update, right?

[...]

PR #1213, opened 2020-09-25
#1193: fix consumer inclusion logic

artemptushkin:
Currently, consumer inclusions have the wrong logic regarding pact broker providing pacts with names like `Pact between %s (%version) and %provider`, it is fixed and unit-test sample is supplied

### DIFF
--- a/provider/src/main/kotlin/au/com/dius/pact/provider/junitsupport/loader/PactBrokerLoader.kt
+++ b/provider/src/main/kotlin/au/com/dius/pact/provider/junitsupport/loader/PactBrokerLoader.kt
@@ -11,6 +11,7 @@ import au.com.dius.pact.core.model.PactSource
 import au.com.dius.pact.core.pactbroker.ConsumerVersionSelector
 import au.com.dius.pact.core.pactbroker.IPactBrokerClient
 import au.com.dius.pact.core.pactbroker.PactBrokerClient
+import au.com.dius.pact.core.support.contains
 import au.com.dius.pact.core.support.expressions.DataType
 import au.com.dius.pact.core.support.expressions.ExpressionParser.parseExpression
 import au.com.dius.pact.core.support.expressions.ExpressionParser.parseListExpression
@@ -200,7 +201,9 @@ open class PactBrokerLoader(
 
       if (pactBrokerConsumers.isNotEmpty()) {
         val consumerInclusions = pactBrokerConsumers.flatMap { parseListExpression(it, resolver) }
-        consumers = consumers.filter { consumerInclusions.isEmpty() || consumerInclusions.contains(it.name) }
+        if (consumerInclusions.isNotEmpty()) {
+          consumers = consumers.filter { consumer -> consumerInclusions.any { inclusion -> consumer.name.contains(inclusion) } }
+        }
       }
 
       return consumers.map { pactReader.loadPact(it, pactBrokerClient.options) }
--- a/provider/junit/src/test/groovy/au/com/dius/pact/provider/junit/loader/PactBrokerLoaderSpec.groovy
+++ b/provider/junit/src/test/groovy/au/com/dius/pact/provider/junit/loader/PactBrokerLoaderSpec.groovy
@@ -538,6 +538,27 @@ class PactBrokerLoaderSpec extends Specification {
     result.size() == 3
   }
 
+  def 'Loads pacts for provided consumers with the specified tags despite on long pact name from the broker'() {
+    given:
+        consumers = ['foo-consumer']
+        tags = ['demo']
+        def expected = [
+                new PactBrokerResult('Pact between foo-consumer and baz-provider', '', '', [], [], false, 'demo', false),
+                new PactBrokerResult('Pact between unknown-consumer and baz-provider', '', '', [], [], false, 'demo', false),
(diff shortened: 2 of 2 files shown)


anto-ac:
@uglyog @pendsley It would be good if you could review this when you get a chance. You're probably more familiar with this area than me.

provider/src/main/kotlin/au/com/dius/pact/provider/junitsupport/loader/PactBrokerLoader.kt
```diff
@@ -200,7 +201,9 @@ open class PactBrokerLoader(
 
       if (pactBrokerConsumers.isNotEmpty()) {
         val consumerInclusions = pactBrokerConsumers.flatMap { parseListExpression(it, resolver) }
-        consumers = consumers.filter { consumerInclusions.isEmpty() || consumerInclusions.contains(it.name) }
+        if (consumerInclusions.isNotEmpty()) {
+          consumers = consumers.filter { consumer -> consumerInclusions.any { inclusion -> consumer.name.contains(inclusion) } }
```
pendsley, line 205: The risk with this is that there are a lot of words in the name that could contain the specified consumer name. Right now, all results are being filtered out, but this could make it match so not enough are being filtered out.

As was pointed out in the issue, `consumer.name` here will be something like
> Pact between consumer (whatever-the-consumer-version-is) and provider

A contrived example would be if I have a provider (my-provider) that serves two consumers (my-consumer and my-consumer-legacy)

> Pact between my-consumer-legacy (123abc) and my-provider
> Pact between my-consumer (456def) and my-provider

In this case, if you had the consumer property set to `my-consumer`, it would still run verifications for `my-consumer-legacy`

artemptushkin: I agree and thought about that, decided to discuss it here.

As I see this endpoint `/pacts/provider/billing/latest` returns:
```json
"pb:pacts": [
            {
                "href": "https://pact-broker.everon.dev/pacts/provider/billing/consumer/dataflow/version/1231f00fea1c41259472b9be79fe69bca0bd178a",
                "title": "Pact between dataflow (1231f00fea1c41259472b9be79fe69bca0bd178a) and billing",
                "name": "dataflow"
            }
        ]
```
And it is okay, we can take the name from there

but this `pacts/provider/billing/for-verification` returns:
```json
"_links": {
                    "self": {
                        "href": "https://pact-broker.everon.dev/pacts/provider/billing/consumer/dataflow/pact-version/af8018d15849cd333429faa7404171d2b4ab588d",
                        "name": "Pact between dataflow (1231f00fea1c41259472b9be79fe69bca0bd178a) and billing"
                    }
                }
```
I see two options:

1. if pact broker will align `pacts/provider/billing/for-verification` with `/pacts/provider/billing/latest` then it will be okay and no complicated filters will be needed, because we'll have `name` and `title`.

2. `pact-jvm` changes the logic of using endpoints and here I'm not aware of the reasoning behind it

I'd made a rush maybe, but now I see this is a better solution, please consider @pendsley @anto-ac
If you see that it should be changed on Pact broker side I'd appreciate some reference in this task to PactBroker one

anto-ac:
cc @bethesque

uglyog:
There is a fix I've been waiting for in the broker project (https://github.com/pact-foundation/pact_broker/issues/350). Ideally, we shouldn't rely on the response to filter the pacts, but send the consumers we want with the request. There was an issue that you can't use the selectors without a tag. This has now been fixed.

We should close this PR (sorry for the wasted effort), and add the selectors based on the filters that are provided.

artemptushkin:
@uglyog no problem, I wanted to get better this part still. I'd apply the relevant fix but I don't know how that pact broker changes are related. Anyway, will be waiting in the issue thread

provider/src/main/kotlin/au/com/dius/pact/provider/junitsupport/loader/PactBrokerLoader.kt
bethesque: I can add the consumer and provider names into the 'pacts for verification' response. I didn't realise anyone was filtering on the title!

artemptushkin: it uses `name`, but at `/for-verification` name value looks like _title_ and the `title` is missing

I feel like we need pacts for pact client libs :)

bethesque: Indeed. Unfortunately, this particular bug arises not because of a format misunderstanding between pact-jvm and the Pact Broker, but because there is a change in where the responsibility lies for filtering by consumer. 

For the old API, the client was doing it. For the new API, the Pact Broker is doing it.

The correct thing to do with the consumer name filter is to pass the name to the Pact Broker via the selectors, and filter on the server side, as this is now supported.

The problem is that there are going to be some people who are still going to be using the old API to fetch the pacts. The consumer filter should only be applied client side if the old API was used to fetch the pacts. That is going to have to be passed in to the filtering code somehow.

bethesque:
https://github.com/pact-foundation/pact_broker/issues/350 is fixed now btw @uglyog

uglyog:
Closing this PR as the fix from 677cc2d should address the problem

Issue #1215, opened 2020-09-28
Request for upgrading netty version in gradle.properties


gayatreemishra:
Hi @rholshausen 

We are facing issue while downloading pact jvm dependencies as it is having a high severity vulnerability (https://snyk.io/vuln/SNYK-JAVA-IONETTY-564897). Could you please help us in upgrading io.netty:netty-codec to version 4.1.46.Final or higher.

PR #1216, opened 2020-09-29
Update gradle.properties

rejeeshg:
addressing Request for upgrading netty version in gradle.properties #1215


uglyog:
Might as well upgrade to the latest, which is 4.1.52.Final

PR #1256, opened 2020-11-25
Configurable HttpClientFactory

Sabartius:
Hey,
the current implementation of HttpTargets made it quite hard to change the Implementation of the httpClient used for the Tests. Currently we use a custom subclass with a lot of copied code, just to change the httpClientFactory, so i thought as the rest of the code is already nicely prepared for extension (interface for the client-factory i.e), why not let the factory be customizable too?

Hope the code is alright, this is my first edit of any kotlin-code ever 🙈


uglyog:
You're edit of the Kotlin code looks fine to me :-D

PR #1276, opened 2020-12-16
fix: missing failures when merging verification results

PhilHardwick:
When failures and success results had the same interactionId, the failures would be
overwritten by an empty list. This fix sees if there's a failure with the same
interactionId and uses the list of failues to map to the interactionId, otherwise
we just use the empty list.


anto-ac:
@PhilHardwick I'm curious... under which circumstances would you have failures and successes for the same `interactionId`?

PhilHardwick:
In my tests this: https://github.com/pact-foundation/pact-jvm/blob/0e339932a5c9dda7d8d725fcf05b6d207b3c0c85/provider/src/main/kotlin/au/com/dius/pact/provider/ProviderVerifier.kt#L533 merges an Ok from status and headers with failures from the body.

uglyog:
Thanks for the PR

PR #1297, opened 2021-01-26
docs: Add docs for LambdaDSL, with minor style fixes

joklek:
Add docs for LambdaDSL, mainly copied from the old method javadocs. Automatic fixes for styling, renamed a few parameters, to better reflect their use

EDIT: Just noticed this https://github.com/pact-foundation/pact-jvm/releases/tag/4_2_0-beta_3 and it seems like the classes I worked on will be replaced with Kotlin :( At least they have their documentation added

### DIFF
--- a/consumer/java8/src/main/java/io/pactfoundation/consumer/dsl/LambdaDsl.java
+++ b/consumer/java8/src/main/java/io/pactfoundation/consumer/dsl/LambdaDsl.java
@@ -1,5 +1,6 @@
 package io.pactfoundation.consumer.dsl;
 
+import au.com.dius.pact.consumer.dsl.DslPart;
 import au.com.dius.pact.consumer.dsl.PactDslJsonArray;
 import au.com.dius.pact.consumer.dsl.PactDslJsonBody;
 import au.com.dius.pact.core.model.matchingrules.MaxTypeMatcher;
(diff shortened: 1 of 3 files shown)


consumer/java8/src/main/java/io/pactfoundation/consumer/dsl/LambdaDslJsonArray.java
```diff
@@ -448,81 +568,100 @@ public LambdaDslJsonArray eachArrayWithMinLike(int numberExamples, Integer size,
...
+     *
+     * @param expression Datetime expression
+     * @param format     Datetime format to use for values
+     */
+    public LambdaDslJsonArray datetimeExpression(String expression, String format) {
+        pactArray.datetimeExpression(expression, format);
+        return this;
+    }
 
     public DslPart build() {
```
joklek, line 664: Moved build() down, to be at the bottom of the file

consumer/java8/src/main/java/io/pactfoundation/consumer/dsl/LambdaDsl.java
```diff
@@ -13,13 +14,22 @@
  */
 public class LambdaDsl {
 
+    private LambdaDsl() {
```
joklek, line 17: Make implicit public no-arg constructor private just to be as tidy as possible

uglyog:
@joklek these classes haven't been changed yet, I think it is still valuable to merge this?

joklek:
Ah, my internal pessimist won without any response :D Thanks for your input, I reopened the PR

PR #1302, opened 2021-02-06
FIX: Support setting consumer and provider pacticipant names from env…

cjr125:
…ironment variables in consumer contract tests

### DIFF
--- a/consumer/junit5/src/main/kotlin/au/com/dius/pact/consumer/junit5/PactConsumerTestExt.kt
+++ b/consumer/junit5/src/main/kotlin/au/com/dius/pact/consumer/junit5/PactConsumerTestExt.kt
@@ -149,7 +149,8 @@ data class ProviderInfo @JvmOverloads constructor(
 
   companion object {
     fun fromAnnotation(annotation: PactTestFor): ProviderInfo =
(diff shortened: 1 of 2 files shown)


consumer/junit5/src/main/kotlin/au/com/dius/pact/consumer/junit5/PactConsumerTestExt.kt
```diff
@@ -149,7 +149,8 @@ data class ProviderInfo @JvmOverloads constructor(
 
   companion object {
     fun fromAnnotation(annotation: PactTestFor): ProviderInfo =
-      ProviderInfo(annotation.providerName, annotation.hostInterface, annotation.port,
+      ProviderInfo(parseExpression(annotation.providerName, DataType.STRING)?.toString() ?: "default",
```
uglyog, line 152: default value should be `annotation.providerName`

uglyog (changes requested):
The default values should just be the original annotation values

consumer/junit5/src/main/kotlin/au/com/dius/pact/consumer/junit5/PactConsumerTestExt.kt
cjr125: done

PR #1304, opened 2021-02-07
Support for x-thrift as json

keeping-it-up:
This change allows to use JsonBodyMatcher for application/x-thrift protocol with json content.
A system property pact.content_type.override.application/x-thrift is added, if its provided with value of json then JsonBodyMatcher will be used to evaluate expression present in contract.
If this property is not provided (default case) then pact matcher will behave as it used to.


eleco:
Use case for this pr is to allow overriding of the default content type mapping logic when the payload is json but the content type is not recognised as one of the standard json content types. For instance if for legacy reason a payload which is json and a content-type "application/x-thrift".

Issue #1316, opened 2021-02-26
`null` value from the Provider response matches regex and incorrectly passes the test


ryanlevell:
Java version: 8
Test framework: Junit5
Surefire plugin: 2.22.0
Pact version: 4.1.17

I have a Pact with an alphanumeric regex for the `meta` body field, `[a-zA-Z0-9]+`:
```
DslPart body = new PactDslJsonBody()
  .stringMatcher("id", "\\d+", "123")
  .stringMatcher("meta", "[a-zA-Z0-9]+", "abc123");
```

If the Provider responds with `meta=null`:
```
{
  "id": 1,
  "content": null,
  "meta": null
}
```

Then test passes unexpectedly. I assume because `null` technically matches the regex.
My expectation is that `null` would fail before getting to the matcher logic.

Actual: Test passes.
Expected: Test fails.

Java logs:
```
Verifying a pact between consumer1 and provider1
  [Using File /target/pacts/consumer1-provider1.json]
  Given test state
  ExampleJavaConsumerPactTest test interaction
....
2021-02-26 08:59:15.523 DEBUG org.apache.http.headers : http-outgoing-0 >> GET /greeting HTTP/1.1
2021-02-26 08:59:15.523 DEBUG org.apache.http.headers : http-outgoing-0 >> Content-Type: application/json
2021-02-26 08:59:15.523 DEBUG org.apache.http.headers : http-outgoing-0 >> Host: localhost:56138
....
2021-02-26 08:59:15.671 DEBUG org.apache.http.wire : http-outgoing-0 << "Content-Type: application/json[\r][\n]"
2021-02-26 08:59:15.671 DEBUG org.apache.http.wire : http-outgoing-0 << "{"id":1,"content":null,"meta":null}[\r][\n]"
....
returns a response which
  has status code 200 (OK)
  has a matching body (OK)
```

Pact Body section:
```
"body": {
  "$.id": {
    "matchers": [
      {
        "match": "regex",
        "regex": "\\d+"
      }
    ],
    "combine": "AND"
  },
  "$.meta": {
    "matchers": [
      {
        "match": "regex",
        "regex": "[a-zA-Z0-9]+"
      }
    ],
    "combine": "AND"
  }
}
```

PR #1317, opened 2021-02-27
Added Check for JsonValue.Null for matchRegex

ryanlevell:
See #1316 - Tests may incorrectly pass when the Provider sends a `null` required JSON field due to the regex matcher.

The reason for this is that `JsonValue.Null` is serialized into the String `"null"`. Then it is matched against the regex, `"null".matches("[a-z0-9]+")`, and will incorrectly result in a successful match. `JsonValue.Null` should never match a regex.


uglyog:
Thanks for the PR.

Issue #1368, opened 2021-05-18
ClasscastException when validating v4 PACT (SynchronousHttp cannot be cast to class RequestResponseInteraction)


gtudan:
I'm trying to validate a v4 PACT with the junit5 provider-tests but get an exception:

```
1) Verifying a pact between push-tan-sdk-rest and push-notification-fs-rest - registration

    1.1)       class au.com.dius.pact.core.model.V4Interaction$SynchronousHttp cannot be cast to class au.com.dius.pact.core.model.RequestResponseInteraction (au.com.dius.pact.core.model.V4Interaction$SynchronousHttp and au.com.dius.pact.core.model.RequestResponseInteraction are in unnamed module of loader 'app')
	at au.com.dius.pact.provider.junit5.PactVerificationContext.verifyInteraction(PactVerificationContext.kt:62)
```

I'm running the latest version (4.2.5) with Junit5.

Here's the PACT:

```json
{
  "provider": {
    "name": "push-notification-fs-rest"
  },
  "consumer": {
    "name": "push-tan-sdk-rest"
  },
  "interactions": [{
      "_id": "d17c5a35c94b0153efa22b40b38a7a31012f587d",
      "type": "Synchronous/HTTP",
      "description": "registration",
      "request": {
        "method": "POST",
        "path": "/push-notification-functional-service/device",
        "query": {
        },
        "headers": {
          "Content-Type": "application/json"
        },
        "body": {
          "deviceId": "4de4adb9-21ee-47e3-aeb4-8cf8ed6c109a",
          "deviceToken": "string",
          "operatingSystem": "iOS"
        }
      },
      "response": {
        "status": 204,
        "headers": {
        },
        "body": null
      },
      "providerStates": [{
          "name": "new device",
          "params": {
          }
        }]
    }],
  "metadata": {
    "pactSpecification": {
      "version": "4.0"
    },
    "pact-dart": {
      "version": "0.0.1"
    }
  },
  "createdAt": "2021-05-18T13:31:24+00:00"
}
```

uglyog:
Linking to #1376

PR #1369, opened 2021-05-19
Fix: Allow verifying v4-interactions with junit5

gtudan:
This is a quick and dirty solution for #1368

I don't know if there is a plan on how to handle the v3 vs. v4 models in the provider tests, so please feel free to close this if there's already something in progress. Or point me in the right direction :-)


uglyog:
Thanks for the PR. There is going to be lots of places which have the concrete classes and need to be changed.

uglyog:
Linking to #1376

Issue #1405, opened 2021-07-20
pact:can-i-deploy retryInterval/retriesWhenUnknown not working as expected


bawoodruff:
Per the Slack channel conversation:

When I have an unknown version of the pact pushed to the pact broker, I expect that the retry process would occur with the maven plugin until a result is posted or the retry period has expired.  Instead, the plugin returns a failure case immediately. Our system is using the OSS pactbroker 2.79.0.

The pact CLI works as expected in this scenario but maven is better suited for our CI configuration.  Logging from -X for pact:can-i-deploy is below.

01:15:25 [DEBUG] -----------------------------------------------------------------------
01:15:25 [DEBUG] Goal:          au.com.dius.pact.provider:maven:4.2.7:can-i-deploy (default-cli)
01:15:25 [DEBUG] Style:         Regular
01:15:25 [DEBUG] Configuration: <?xml version="1.0" encoding="UTF-8"?>
01:15:25 <configuration>
01:15:25   <latest default-value="">true</latest>
01:15:25   <pactBrokerAuthenticationScheme default-value="basic">basic</pactBrokerAuthenticationScheme>
01:15:25   <pactBrokerPassword>${pactbroker.auth.password}</pactBrokerPassword>
01:15:25   <pactBrokerServerId>${pact.broker.serverId}</pactBrokerServerId>
01:15:25   <pactBrokerToken>${pact.broker.token}</pactBrokerToken>
01:15:25   <pactBrokerUrl>https://redacted</pactBrokerUrl>
01:15:25   <pactBrokerUsername>${pactbroker.auth.username}</pactBrokerUsername>
01:15:25   <pacticipant>PetstoreService</pacticipant>
01:15:25   <pacticipantVersion>${pacticipantVersion}</pacticipantVersion>
01:15:25   <retriesWhenUnknown default-value="0">10</retriesWhenUnknown>
01:15:25   <retryInterval default-value="10">300</retryInterval>
01:15:25   <settings default-value="${settings}"/>
01:15:25   <to default-value="">${pact.tag}</to>
01:15:25 </configuration>
01:15:25 [DEBUG] =======================================================================
01:15:25 [INFO] 
01:15:25 [INFO] --- maven:4.2.7:can-i-deploy (default-cli) @ bryanwconsumer01 ---
...
[...]

PR #1406, opened 2021-07-21
fix: retry condition for the pactbroker unknown response #1405

bawoodruff:


uglyog:
Damn, that is a perfect example of what happens when you have a gap in your tests!

uglyog:
Actually, there was a test, but it mocked things out to match the implementation

PR #1411, opened 2021-08-08
feat: add support for verification type RESPONSE_FACTORY. Fixes #1379

### LINKED ISSUE #1379: New verification type: FUNCTION
Currently there exist two verification types:
- `PactVerification.REQUEST_RESPONSE` -- verifies an HTTP response from a request
- `PactVerification.ANNOTATED_METHOD` -- can be used to verify either an HTTP response, or a message and metadata

I would like to propose a third verification type, `PactVerification.FUNCTION`, which would be similar to `PactVerification.ANNOTATED_METHOD` except that:
- only a single function is required, it would be a factory of `String` -> `Any` (but really, it would return specific types that the library already understands, like `MessageAndMetadata`).
- the function would be provided explicitly to `ProviderVerifier` instead of via scanning the classpath for annotated methods
- the function would not be invoked by reflection (e.g. `java.lang.reflect.Method.invoke()`), it would just be called normally like any other callback method in the `ProviderVerifier`. this means the method scope would be the same as any method in the test class instead of having to provide a specific instance for invocation.

This could be used for both messages and request/response, although the return types would need to vary depending on the interaction type. Probably the return value can be the same as the return values that are currently supported by the `PactVerification.ANNOTATED_METHOD` type.

My guess is the easiest way to implement this is to have another bifurcation in `ProviderVerifier.verifyInteraction` which would call a different method if the verification type is `FUNCTION`, for example `verifyResponseByFunction`, and define a new property on the `ProviderVerifier` that would contain the available function to call (`verificationFunction: Function<String, Any>`, or something like that). There can probably be some refactoring around `verifyResponseByInvokingProviderMethods` and `verifyMessage` to split common functionality into common methods, since mostly the only change is how to call the method and get the return value.

kflorence:


kflorence:
@uglyog i'm not sure if the test coverage is thorough enough (or even correct -- i'm not super familiar with groovy), and I also wanted to get your input on naming -- but otherwise I think this is mostly good to go for the initial iteration.

kflorence:
It would also be good to backport this change to the `4.1.x` branch for JDK 8 support

JapuDCret:
If this is accepted, then the usage should be documented in https://github.com/pact-foundation/docs.pact.io

kflorence:
@JapuDCret yes, it should be documented somewhere... however, it seems the documentation is already pretty out of date https://docs.pact.io/implementation_guides/jvm/provider -- there isn't even mention of the annotated method of verification. Maybe it would make more sense to update the documentation once this feature lands in one of the other modules, like JUnit.

FYI the primary intent of this was to make the library easier to use in Scala with https://github.com/jbwheatley/pact4s

provider/src/main/kotlin/au/com/dius/pact/provider/ProviderVerifier.kt
```diff
@@ -349,6 +361,57 @@ open class ProviderVerifier @JvmOverloads constructor (
...
+          failures,
+          pending
+        )
+      } else {
+        val expectedResponse = (interaction as SynchronousRequestResponse).response
+        val response = factory.apply(interaction.description) as Map<String, Any>
+        val actualResponse = ProviderResponse(
+          response["statusCode"] as Int,
+          response["headers"] as Map<String, List<String>>,
+          ContentType.UNKNOWN,
```
uglyog, line 391: Could the factory function not also supply the content type?

provider/src/test/groovy/au/com/dius/pact/provider/ProviderVerifierSpec.groovy
```diff
@@ -759,4 +759,79 @@ class ProviderVerifierSpec extends Specification {
...
+            false
+    )
+
+    then:
+    result instanceof VerificationResult.Ok
+  }
+
+  def 'verifyResponseByFactory returns an error result if the factory method fails with an exception'() {
+    given:
+    ProviderInfo provider = new ProviderInfo('Test Provider')
```
uglyog, line 813: Is this test passing because the factory function is not set? 
If so, it would be better to be explicit and have it setup a function that throws an exception on invocation

uglyog (changes requested):

provider/src/main/kotlin/au/com/dius/pact/provider/ProviderVerifier.kt
kflorence: @uglyog i copied this logic from here https://github.com/pact-foundation/pact-jvm/blob/master/provider/src/main/kotlin/au/com/dius/pact/provider/ProviderVerifier.kt#L327

Do we want to update that in both places?

provider/src/test/groovy/au/com/dius/pact/provider/ProviderVerifierSpec.groovy
kflorence: that is correct. I will make it a more explicit test

provider/src/main/kotlin/au/com/dius/pact/provider/ProviderVerifier.kt
uglyog: No, don't worry about that, I will fix the other place.

uglyog (approved):

kflorence:
@uglyog thanks! sorry for breaking the codenarc tests in my last commit. Is it possible to get this backported into the 4.1.x branch?

uglyog:
Done. Backported.

Issue #1324, opened 2021-03-12
@IgnoreNoPactsToVerify test fails with NotFoundHalResponse when no pacts found 


ceccone:
While running a JUnit5/Spring Pact verification  with `@IgnoreNoPactsToVerify` and `@PactBroker` when there are no published pacts in the pact-broker, the test fails with `NotFoundHalResponse` exception.

The pact libraries we're using:
`testImplementation "au.com.dius.pact.provider:junit5spring:4.2.1"`

\*EDIT: works with version `4.1.8` but not with `4.1.9` onwards.
My best guess is #1225 broke it.

The complete stacktrace:
```
au.com.dius.pact.core.pactbroker.NotFoundHalResponse: No HAL document found at path 'http://pact-broker/pacts/provider/foo-provider/for-verification'
	at au.com.dius.pact.core.pactbroker.HalClient.handleHalResponse(HalClient.kt:299)
	at au.com.dius.pact.core.pactbroker.HalClient.access$handleHalResponse(HalClient.kt:141)
	at au.com.dius.pact.core.pactbroker.HalClient$postJson$3.invoke(HalClient.kt:474)
	at au.com.dius.pact.core.support.KotlinLanguageSupportKt.handleWith(KotlinLanguageSupport.kt:38)
	at au.com.dius.pact.core.pactbroker.HalClient.postJson(HalClient.kt:472)
	at au.com.dius.pact.core.pactbroker.PactBrokerClient$fetchPactsUsingNewEndpoint$2.invoke(PactBrokerClient.kt:256)
	at au.com.dius.pact.core.support.KotlinLanguageSupportKt.handleWith(KotlinLanguageSupport.kt:38)
	at au.com.dius.pact.core.pactbroker.PactBrokerClient.fetchPactsUsingNewEndpoint(PactBrokerClient.kt:255)
	at au.com.dius.pact.core.pactbroker.PactBrokerClient.fetchConsumersWithSelectors(PactBrokerClient.kt:214)
	at au.com.dius.pact.provider.junitsupport.loader.PactBrokerLoader.loadPactsForProvider(PactBrokerLoader.kt:218)
	at au.com.dius.pact.provider.junitsupport.loader.PactBrokerLoader.load(PactBrokerLoader.kt:116)
[...]

PR #1421, opened 2021-08-23
feat: add handling for 404 responses in Maven plugin #1419

### LINKED ISSUE #1419: Pact Maven Plugin failIfNoPactsFound fails with NotFoundHalResponse when no pacts found/provider unknown
# The issue

In the [implementation guide for Pacts Maven plugin](https://docs.pact.io/implementation_guides/jvm/provider/maven/#not-failing-the-build-if-no-pact-files-are-found) it is mentioned, that you can use `failIfNoPactsFound` to ignore if there are no pacts defined for a given provider. But when the provider is unknown to the broker, the broker will not return an empty list but a 404:

```bash
HTTP/1.1 404 Not Found
Content-Type: application/hal+json;charset=utf-8

{"error":"No provider with name 'MyEventProvider' found"}
```

This causes the plugin to throw an exception: `Execution verify-pact-contracts of goal au.com.dius.pact.provider:maven:4.2.8:verify failed: Call to fetch pacts from Pact Broker failed with an exception: No HAL document found at path 'http://localhost:9292/pacts/provider/MyEventProvider/for-verification'`.

I suspect that the `PactProviderMojo.loadPactsFromPactBroker()` cannot handle a 404 response by the broker (see [here](https://github.com/pact-foundation/pact-jvm/blob/master/provider/maven/src/main/kotlin/au/com/dius/pact/provider/maven/PactProviderMojo.kt#L137-L192)).
I think the same change as for the `@IgnoreNoPactsToVerify` needs to be done (see #1324).

# My setup

```xml
            <plugin>
                <groupId>au.com.dius.pact.provider</groupId>
                <artifactId>maven</artifactId>
                <version>4.2.8</version>
                <configuration>
                    <pactBrokerUrl>http://localhost:9292</pactBrokerUrl>
                    <projectVersion>0.1-SNAPSHOT</projectVersion>
                    <trimSnapshot>true</trimSnapshot> <!-- the pact broker has issues parsing SNAPSHOT versions -->
                    <skipPactPublish>false</skipPactPublish>
                    <failIfNoPactsFound>false</failIfNoPactsFound> <!-- FIXME: set to true, once the initial pacts are available -->
                    <tags>
                        <tag>dev</tag>
                    </tags>
                    <systemPropertyVariables>
                        <pact.verifier.publishResults>true</pact.verifier.publishResults>
                    </systemPropertyVariables>
                    <serviceProviders>
                        <serviceProvider>
                            <name>MyEventProvider</name>
                            <verificationType>ANNOTATED_METHOD</verificationType>
                        </serviceProvider>
                    </serviceProviders>
                </configuration>
            </plugin>
```

# Misc

For some providers, the broker seems to "know"(?) them and returns a 200 with an empty list, this the plugin can handle:

```bash
HTTP/1.1 200 OK
Content-Type: application/hal+json;charset=utf-8

{"_embedded":{"pacts":[]},"_links":{"self":{"href":"http://localhost:9292/pacts/provider/MyOtherEventProvider/for-verification","title":"Pacts to be verified"}}}
```

# Expected behaviour

That the `failIfNoPactsFound` works like `@IgnoreNoPactsToVerify`, in that it can also ignore a 404 being returned by the pact broker. Or the pact broker not returning a 404.

JapuDCret:
This PR is based on #1324 and fixes #1419.

### DIFF
--- a/provider/maven/src/main/kotlin/au/com/dius/pact/provider/maven/PactProviderMojo.kt
+++ b/provider/maven/src/main/kotlin/au/com/dius/pact/provider/maven/PactProviderMojo.kt
@@ -1,6 +1,8 @@
 package au.com.dius.pact.provider.maven
 
 import au.com.dius.pact.core.pactbroker.ConsumerVersionSelector
+import au.com.dius.pact.core.pactbroker.NotFoundHalResponse
+import au.com.dius.pact.core.support.handleWith
 import au.com.dius.pact.core.support.toUrl
 import au.com.dius.pact.provider.ConsumerInfo
 import au.com.dius.pact.provider.IConsumerInfo
@@ -12,6 +14,7 @@ import au.com.dius.pact.provider.ProviderVerifier
 import au.com.dius.pact.provider.ProviderVersion
 import au.com.dius.pact.provider.VerificationResult
 import au.com.dius.pact.provider.reporters.ReporterManager
+import com.github.michaelbull.result.getOrElse
 import org.apache.maven.plugin.MojoFailureException
 import org.apache.maven.plugins.annotations.Mojo
 import org.apache.maven.plugins.annotations.Parameter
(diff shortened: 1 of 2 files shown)


JapuDCret:
I did not test these changes locally (other than the unit tests), as I'm unable to locally build the project.

PR #1443, opened 2021-09-18
feat: add support for non-file parts in the multipart requests

kmochocki:
Hello

While working on multipart I couldn't call my API accepting text and file parameters. After some debugging I noticed that during request creation at the provider the parts are always treated as a files. Further investigation lead me to the conclusion that the non-file parts should be treated as the parameters (`org.apache.commons.fileupload.FileUploadBase#parseRequest(org.apache.commons.fileupload.RequestContext` + `org.apache.commons.fileupload.FileUploadBase.FileItemIteratorImpl#findNextItem` + `org.springframework.web.multipart.commons.CommonsFileUploadSupport#parseFileItems` )

### DIFF
--- a/provider/junit5spring/src/main/kotlin/au/com/dius/pact/provider/spring/junit5/MockMvcTestTarget.kt
+++ b/provider/junit5spring/src/main/kotlin/au/com/dius/pact/provider/spring/junit5/MockMvcTestTarget.kt
@@ -98,7 +98,7 @@ class MockMvcTestTarget @JvmOverloads constructor(
           val name = StringUtils.defaultString(contentDisposition.getParameter("name"), "file")
           val filename = contentDisposition.getParameter("filename").orEmpty()
           if (filename.isEmpty()) {
-            multipartRequest.param(name, FileCopyUtils.copyToString(bodyPart.inputStream.bufferedReader()))
+            multipartRequest.part(MockPart(name, FileCopyUtils.copyToByteArray(bodyPart.inputStream)))
           } else {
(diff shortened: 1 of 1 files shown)


provider/junit5spring/src/main/kotlin/au/com/dius/pact/provider/spring/junit5/MockMvcTestTarget.kt
```diff
@@ -96,7 +97,11 @@ class MockMvcTestTarget @JvmOverloads constructor(
           val contentDisposition = ContentDisposition(bodyPart.getHeader("Content-Disposition").first())
           val name = StringUtils.defaultString(contentDisposition.getParameter("name"), "file")
           val filename = contentDisposition.getParameter("filename").orEmpty()
-          multipartRequest.file(MockMultipartFile(name, filename, bodyPart.contentType, bodyPart.inputStream))
+          if (filename.isEmpty()) {
+            multipartRequest.param(name, FileCopyUtils.copyToString(bodyPart.inputStream.bufferedReader()))
```
kmochocki, line 101: Tbh it should be:
```suggestion
            multipartRequest.part(MockPart(name, FileCopyUtils.copyToByteArray(bodyPart.inputStream)))
```

But there is a problem with parsing parts in `MockMvc` in used version. It is already fixed (https://github.com/spring-projects/spring-framework/issues/26400) but requires library to be updated.

uglyog: I have upgraded the Spring libraries to 5.3.9

PR #1455, opened 2021-09-30
fix(PactBrokerClient): Send `includePendingStatus=false` when enablePending is set to false

TimothyJones:


uglyog:
A test for that would be nice

uglyog (approved):

uglyog:
Needs a test here: https://github.com/pact-foundation/pact-jvm/blob/master/core/pactbroker/src/test/groovy/au/com/dius/pact/core/pactbroker/PactBrokerClientSpec.groovy#L463

TimothyJones:
No problem. I'll add one in a moment.

Issue #1465, opened 2021-10-13
Add support for branches in consumer version selectors


bethesque:
The Pact Broker now supports branches as first class entities. You can read more about this here: https://docs.pact.io/pact_broker/branches

## Description

Please add support for specifying branch related properties in the consumer version selectors that are sent to the Pact Broker when requesting pacts to verify. Please do not do any verification of the consumer version selectors on the client side - the validation rules are subject to change. Just ensure that any error response is displayed to the user.

* For implementations that wrap the pact-ruby-standalone, update to the latest standalone version
* For implementations that use the rust implementation, update to the latest rust version.
* Add a String `branch` property to the consumer version selector (if there is a domain model for this).
* Add a Boolean `mainBranch` (or `main_branch` for snake case languages) property consumer version selector (if there is a domain model for this).
* Add a Boolean `matchingBranch` (or `matching_branch` for snake case languages) property consumer version selector (if there is a domain model for this).
* Expose and document the `branch`, `mainBranch` (or `main_branch` for snake case languages) and `matchingBranch` (or `matching_branch`) properties in the user facing API.
* Pass the `branch`, `mainBranch` and `matchingBranch` (must be camelcase) through to the relevant implementation (ruby and/or rust)

## Verifying the changes

* Publish test pacts to the test broker

```
export PACT_BROKER_BASE_URL="https://test.pact.dius.com.au"
export PACT_BROKER_USERNAME="dXfltyFMgNOFZAxr8io9wJ37iUpY42M"
export PACT_BROKER_PASSWORD="O5AIZWxelWbLvqMd8PkAVycBJh2Psyg1"

docker run --rm  \
  -e PACT_BROKER_BASE_URL  \
  -e PACT_BROKER_USERNAME  \
  -e PACT_BROKER_PASSWORD  \
  pactfoundation/pact-cli:0.50.0.14 \
  publish \
  /pact/example/pacts \
  --consumer-app-version fake-git-sha-for-demo-$(date +%s) \
  --branch main

docker run --rm  \
  -e PACT_BROKER_BASE_URL  \
  -e PACT_BROKER_USERNAME  \
  -e PACT_BROKER_PASSWORD  \
  pactfoundation/pact-cli:0.50.0.14 \
  publish \
  /pact/example/pacts \
  --consumer-app-version fake-git-sha-for-demo-$(date +%s) \
  --branch feat/x
```

* Using your pact client library, verify `{ "mainBranch": true }`
  * You should receive the main pact
[...]

PR #1471, opened 2021-10-26
feat: Add support for configuring consumer version selectors from cli as

RadekKoubsky:
raw json

Based on discussion in #1465, I have implemented a solution that bypasses the actual typed selectors implementation by passing raw_json/untyped selectors configuration directly from system property to the pact broker. This can be used as a workaround for not yet implemented new broker api in typed code.

I tested it on the latest pact broker:
Publish pacts
`pact-broker publish --consumer-app-version 1 --broker-base-url http://localhost:9292 ./pacts --branch master`
Run verification tests and publish results
`mvn clean test -Dpactbroker.url=http://localhost:9292 -Dpactbroker.consumerversionselectors.rawjson=[{\"mainBranch\":true}] -Dpact.provider.version=1 -Dpact.provider.tag=master -Dpact.verifier.publishResults=true -Dpactbroker.enablePending=true -Dpactbroker.providerTags=master`

Results were published using tag `master`, however, due to the automatic migration property, the branch `master` was also created in broker.

See screenshot of the matrix:
selectors_with_branches_test

### DIFF
--- a/core/pactbroker/src/main/kotlin/au/com/dius/pact/core/pactbroker/PactBrokerClient.kt
+++ b/core/pactbroker/src/main/kotlin/au/com/dius/pact/core/pactbroker/PactBrokerClient.kt
@@ -255,7 +255,12 @@ open class PactBrokerClient(
       else -> null
     }
     return if (pactsForVerification != null) {
-      fetchPactsUsingNewEndpoint(selectors, enablePending, providerTags, includeWipPactsSince, halClient, pactsForVerification, providerName)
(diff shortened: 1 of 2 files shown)


uglyog:
I'm going to merge this, but it needs to be documented. Unfortunately this means it needs to be in all the provider project READMEs: Gradle, Maven, JUnit, Spring, JUnit 5, JUnit 5 + Spring, Lein because people will only read the docs for the framework they are using.

PR #1473, opened 2021-10-29
feat: Add support for TestClient as a test target.

waikontse:
Added support for TestClient as a test target.
Refactored the code where WebFluxTarget and WebTestClientTarget shares a common base class WebFluxBasedTestTarget.

### DIFF
--- a/provider/junit5spring/src/main/kotlin/au/com/dius/pact/provider/spring/junit5/WebTestClientTarget.kt
+++ b/provider/junit5spring/src/main/kotlin/au/com/dius/pact/provider/spring/junit5/WebTestClientTarget.kt
@@ -1,0 +1,16 @@
+package au.com.dius.pact.provider.spring.junit5
+
+import au.com.dius.pact.core.model.Interaction
+import au.com.dius.pact.core.model.SynchronousRequestResponse
+import au.com.dius.pact.core.model.generators.GeneratorTestMode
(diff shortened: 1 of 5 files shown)


provider/junit5spring/src/main/kotlin/au/com/dius/pact/provider/spring/junit5/WebTestClientTarget.kt
```diff
@@ -0,0 +1,16 @@
...
+import au.com.dius.pact.core.model.generators.GeneratorTestMode
+import org.springframework.test.web.reactive.server.WebTestClient
+
+class WebTestClientTarget(private val webTestClient: WebTestClient) : WebFluxBasedTestTarget {
+    override fun prepareRequest(interaction: Interaction, context: MutableMap<String, Any>): Pair<WebTestClient.RequestHeadersSpec<*>, WebTestClient> {
+        if (interaction is SynchronousRequestResponse) {
+            val request = interaction.request.generatedRequest(context, GeneratorTestMode.Provider)
+            return toWebFluxRequestBuilder(webTestClient, request) to webTestClient
+        }
+        throw UnsupportedOperationException("Only request/response interactions can be used with an MockMvc test target")
```
uglyog, line 14: Error message needs to be updated

uglyog (changes requested):

Issue #1454, opened 2021-09-29
Add support for publishing verification results with branches


bethesque:
This change allows the provider version associated with verification results to be created with a branch.

* Add a "branch" option to the API exposed to the users
* When fetching the pacts for verification, set the branch as `providerVersionBranch` in the request.
* When publishing the verification results, before creating the tags, send a `PUT` using the `pb:branch-version` relation from the provider resource (with params `branch` and `version`), or the `pb:pacticipant-branch-version` relation from the index (with params `pacticipant`, `branch` and `version`).
* If the branch is set, and the broker does not support branches (ie. the relations are not there), raise an appropriate error (or print warning - which ever works best for JVM) indicating that the broker does not support branches, and they should update to Pact Broker version 2.86.0 or later for branch support.
* To verify this is working
  * Verify a pact with the branch property set and publish to Pact Broker 2.86.0 or later
  * Open up the matrix page for the integration
  * You should see a slug with the branch name next to the provider version number

bethesque:
You can use [this feature](https://docs.pact.io/pact_broker/branches#automatic-branch-creation-from-first-tag) to automatically set the branch from the first tag while you're waiting, but as you noted, the consumer version selectors need the branch support added to really make use of them.

PR #1477, opened 2021-11-10
feat: Publish verification results with provider branch

RadekKoubsky:
Hi,
this PR implements #1454 

I tested it on the latest pact broker:
Publish pacts
`pact-broker publish --consumer-app-version 1 --broker-base-url http://localhost:9292 ./pacts --branch master`
Run verification tests and publish results
```
mvn clean test -Dpactbroker.url=http://localhost:9292 -Dpactbroker.consumerversionselectors.rawjson=[{\"mainBranch\":true},{\"deployedOrReleased\":true}] -Dpact.provider.version=1 -Dpact.provider.branch=master -Dpact.verifier.publishResults=true -Dpactbroker.enablePending=true -Dpactbroker.providerTags=master
```

Results were published and provider `master` branch was created in pact broker.

See screenshot of the matrix:
provider_branch

@bethesque 
In the old broker, I used to define selectors with tags this way `-Dpactbroker.consumerversionselectors.tags=master,stage`
In the raw json selectors format, what is an equivalent value to `master,stage` given that `stage` is a no-production environment used with `record-deployment` command?
Based on examples from [documentation](https://docs.pact.io/pact_broker/advanced_topics/consumer_version_selectors/#advanced), I have got something like this:
```
[
   {
      "mainBranch":true
   },
   {
      "deployed":true,
      "environment":"stage"
   }
]
```
or if I want to specify the exact branch
```
[
   {
      "branch":"master"
   },
   {
      "deployed":true,
      "environment":"stage"
   }
]
```
Another use case is "I am in a feature branch and I want to verify my provider build only against stage consumers", is this json correct?
```
[
   {
      "deployed":true,
      "environment":"stage"
   }
]
```
Thanks

### DIFF
--- a/core/pactbroker/src/main/kotlin/au/com/dius/pact/core/pactbroker/PactBrokerClient.kt
+++ b/core/pactbroker/src/main/kotlin/au/com/dius/pact/core/pactbroker/PactBrokerClient.kt
@@ -158,6 +158,20 @@ interface IPactBrokerClient {
   ): Result<Boolean, List<String>>
 
   /**
+   * Publish provider branch to the Pact broker
+   * @param docAttributes Attributes associated with the fetched Pact file
+   * @param tag Provider name
+   * @param branch Provider branch
+   * @param version Provider version
+   */
+  fun publishProviderBranch(
+    docAttributes: Map<String, Any?>,
+    name: String,
+    branch: String,
+    version: String
+  ): Result<Boolean, List<String>>
+
+  /**
    * Publishes the result to the "pb:publish-verification-results" link in the document attributes.
    */
   fun publishVerificationResults(
@@ -565,6 +579,35 @@ open class PactBrokerClient(
     }
   }
 
+  override fun publishProviderBranch(
+    docAttributes: Map<String, Any?>,
+    name: String,
+    branch: String,
+    version: String
+  ): Result<Boolean, List<String>> {
+    try {
+      val halClient = newHalClient()
+        .withDocContext(docAttributes)
+        .navigate(PROVIDER)
+      val result = halClient.putJson(PROVIDER_BRANCH_VERSION, mapOf("version" to version, "branch" to branch), "{}")
+      when (result) {
+        is Ok<*> -> {
+          logger.debug { "Pushed branch $branch for provider $name and version $version" }
+          return Ok(true)
+        }
+        is Err<Exception> -> {
+          logger.error(result.error) { "Failed to push branch $branch for provider $name and version $version" }
+          return Err(listOf("Publishing branch '$branch' failed: ${result.error.message ?: result.error.toString()}"))
+        }
+      }
+
+    } catch (e: NotFoundHalResponse) {
+      val message = "Could not create branch for provider $name, link was missing. It looks like your Pact Broker does not support branches, please update to Pact Broker version 2.86.0 or later for branch support"
+      logger.error(e) { message }
+      return Err(listOf(message))
+    }
+  }
+
   @JvmOverloads
   open fun canIDeploy(
     pacticipant: String,
@@ -616,6 +659,7 @@ open class PactBrokerClient(
     const val BETA_PROVIDER_PACTS_FOR_VERIFICATION = "beta:provider-pacts-for-verification"
     const val PROVIDER = "pb:provider"
     const val PROVIDER_TAG_VERSION = "pb:version-tag"
+    const val PROVIDER_BRANCH_VERSION = "pb:branch-version"
     const val PACTS = "pb:pacts"
     const val UTF8 = "UTF-8"
 
--- a/core/pactbroker/src/main/kotlin/au/com/dius/pact/core/pactbroker/PactBrokerClient.kt
+++ b/core/pactbroker/src/main/kotlin/au/com/dius/pact/core/pactbroker/PactBrokerClient.kt
@@ -146,7 +146,7 @@ interface IPactBrokerClient {
   /**
    * Publish all the tags for the provider to the Pact broker
    * @param docAttributes Attributes associated with the fetched Pact file
-   * @param tag Provider name
+   * @param name Provider name
    * @param tags Provider tags to tag the provider with
    * @param version Provider version
    */
@@ -160,7 +160,7 @@ interface IPactBrokerClient {
   /**
    * Publish provider branch to the Pact broker
    * @param docAttributes Attributes associated with the fetched Pact file
-   * @param tag Provider name
+   * @param name Provider name
    * @param branch Provider branch
    * @param version Provider version
    */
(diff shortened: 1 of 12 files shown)


core/pactbroker/src/main/kotlin/au/com/dius/pact/core/pactbroker/PactBrokerClient.kt
```diff
@@ -157,6 +157,20 @@ interface IPactBrokerClient {
     version: String
   ): Result<Boolean, List<String>>
 
+  /**
+   * Publish provider branch to the Pact broker
+   * @param docAttributes Attributes associated with the fetched Pact file
+   * @param tag Provider name
```
bethesque, line 163: old docs?

provider/src/main/kotlin/au/com/dius/pact/provider/ProviderVerifier.kt
```diff
@@ -844,6 +848,30 @@ open class ProviderVerifier @JvmOverloads constructor (
     }
   }
 
+  private fun reportResults(
+      pact: FilteredPact,
+      result: VerificationResult,
+      client: IPactBrokerClient?
+  ): Result<Boolean, List<String>> {
+    if (providerBranch?.get()?.isNotBlank() == true){
```
bethesque, line 856: This seems to only allow publishing with tags OR a branch. It should allow both (or neither) to be used. The branch should be created before the tags.

bethesque (changes requested):

bethesque:
Thanks for picking this up!

bethesque:
@RadekKoubsky yes, those selectors are correct. If you specify an environment, the "deployed" is assumed to be true, so you can just skip that for brevity if you wish. The best practice selectors are `{ mainBranch: true }, { deployedOrReleased: true }` (meaning you're always forwards and backwards compatible). I'm curious as to why your only want to verify the stage environment on a feature branch. Can you tell me more about that? (Maybe on slack, so as not to derail this issue).

uglyog:
Build has failed due to some static code check issues:
```
MaxLineLength - [reportResults] at /home/runner/work/pact-jvm/pact-jvm/provider/src/main/kotlin/au/com/dius/pact/provider/TestResultAccumulator.kt:108:7
MaxLineLength - [reportResults] at /home/runner/work/pact-jvm/pact-jvm/provider/src/main/kotlin/au/com/dius/pact/provider/TestResultAccumulator.kt:112:7
MaxLineLength - [reportResultsWithBranch] at /home/runner/work/pact-jvm/pact-jvm/provider/src/main/kotlin/au/com/dius/pact/provider/VerificationReporter.kt:45:4
ReturnCount - [stateChange] at /home/runner/work/pact-jvm/pact-jvm/provider/src/main/kotlin/au/com/dius/pact/provider/StateChange.kt:99:16
MaxLineLength - [<anonymous>] at /home/runner/work/pact-jvm/pact-jvm/provider/src/main/kotlin/au/com/dius/pact/provider/StateChange.kt:170:13
MaxLineLength - [pactBrokerClient] at /home/runner/work/pact-jvm/pact-jvm/provider/src/main/kotlin/au/com/dius/pact/provider/ProviderInfo.kt:111:5
```

core/pactbroker/src/main/kotlin/au/com/dius/pact/core/pactbroker/PactBrokerClient.kt
RadekKoubsky: +1, I fixed it in the tags method as well

provider/src/main/kotlin/au/com/dius/pact/provider/ProviderVerifier.kt
RadekKoubsky: Thanks for explanation. I don't know why I thought it should be either "tags" or "branch", the specification says "before creating the tags, send a PUT using the pb:branch-version" :)
I will add PUT request to the branch endpoint just before the call to the tags endpoint and remove the conditional logic.

RadekKoubsky:
I have pushed a new implementation that allows creating both provider branch and tags.

I tested the following use cases:
Create only branch
Create only tags
Create both branch and tags
Do not create any branch and tags, just publish results

@bethesque 
I found unexpected behavior when I was testing this with the pact broker `2.89.0.0` and env vars:
```
      PACT_BROKER_CREATE_DEPLOYED_VERSIONS_FOR_TAGS: false
      PACT_BROKER_USE_FIRST_TAG_AS_BRANCH: false
```

Results were sent correctly when I used selectors `[{mainBranch:true},{deployedOrReleased:true}]`

But when I used `[{mainBranch:true},{environment:stage}]`, I got:
```
au.com.dius.pact.core.pactbroker.RequestFailedException: Request to path 'http://localhost:9292/pacts/provider/book_provider/for-verification' failed with response 400
```

@uglyog 
Hi,
can you please share how can I view what tests failed from build logs, I can only see:
```
Caused by: org.gradle.api.GradleException: There were failing tests. See the report at: file:///home/runner/work/pact-jvm/pact-jvm/provider/build/reports/tests/test/index.html
```
I ran `gradle :provider:clean :provider:build` locally with success (can't get over `PactTest > testPactHttps() FAILED` locally)
Thanks

uglyog:
The build reports are not stored in CI (it hasn't been setup since moving to GH actions). I normally just run the tests locally. But leave it with me, I'll sort it out.

RadekKoubsky:
@uglyog thanks for merging it. When this with #1471 can be released? I would like to migrate our CI/CD workflow from tags to the new model as soon as possible, we plan to release new CI/CD only with the new model, completely abandoning tags.

uglyog:
I've back ported it and released it with all the supported versions: 4.1.x, 4.2.x and 4.3.x

bethesque:
> au.com.dius.pact.core.pactbroker.RequestFailedException: Request to path 'http://localhost:9292/pacts/provider/book_provider/for-verification' failed with response 400

Does the code not print out the body of the response when this happens? If it does not, can you please update it to print out the response. There will be an error message in there that tells you why it is a 400. The user needs to be able to see this.

RadekKoubsky:
@uglyog thanks!

@bethesque 
I have found the problem, I did not have stage environment in broker, that's why sending selectors with stage env `[{mainBranch:true},{environment:stage}]` failed, body of the response describes the problem:
```
{"errors":{"consumerVersionSelectors":["environment with name 'stage' does not exist at index 1"]}}
```
Unfortunately, the [HalClient#handleHalResponse](https://github.com/pact-foundation/pact-jvm/blob/master/core/pactbroker/src/main/kotlin/au/com/dius/pact/core/pactbroker/HalClient.kt#L310) method stores response body in exception, but only the generic message of the exception is displayed to user.

I have to integrate the new broker functionality into our CI/CD, thus I cannot work on it right now. Maybe you can create a new issue.

bethesque:
It's quite important that the user can see that message. I'll create a new issue.

PR #1489, opened 2021-12-07
Specify buildUrl with system property pact.verifier.buildUrl

### LINKED ISSUE #1482: Add a way to specify buildUrl when publishing verification results using JUnit5
I can see that the PactBrokerClient supports (in theory) a buildUrl being present ([here](https://github.com/pact-foundation/pact-jvm/blob/master/core/pactbroker/src/main/kotlin/au/com/dius/pact/core/pactbroker/PactBrokerClient.kt#L446)), but as far as I can tell there's no way to set it when using JUnit (5) to do the publishing of verification results?

It would be really helpful to be able to link from the broker (or even better, in a webhook notification for event type `provider_verification_failed` – but I recognise that's on the broker rather than the JVM library) to the output of the build job where the test failure would be visible.

Presumably this sort of thing would need to be specified with a system property, say `pact.verifier.buildUrl`. I'm not clear how deep in the stack you'd want to do that – would you introduce it within the PactBrokerClient's overloaded `publishVerificationResults` which doesn't have a buildUrl parameter? If that'd be acceptable then I'd happily have a go at raising a PR.

michaelbannister:
buildUrl is an optional value when publishing verification results, but there's currently no way to set it using JUnit or (as far as I can tell) any of the other pact-jvm provider verification publication approaches.

This PR proposes a new system property `pact.verifier.buildUrl` which can be set when publishing verification results to specify the buildUrl value.

See https://docs.pact.io/pact_broker/advanced_topics/api_docs/publish_verification_result/

Fixes #1482


uglyog:
Thanks for the PR.

I also going to back-port this to 4.2/4.1.

PR #1503, opened 2022-01-10
feat: Kafka schema registry messaging support

muirandy:
Messaging supports various mechanisms - WebSocket and Kafka to name but two.

When we use Kafka, and write basic messages, we're good. 

When we write messages that are backed by the Schema Registry, we've got a new problem: the "[magic bytes](https://docs.confluent.io/platform/current/schema-registry/serdes-develop/index.html#wire-format)". 

This PR allows us to use a Schema Registry compliant SerDes for JSON.


core/support/src/main/kotlin/au/com/dius/pact/core/support/json/KafkaSchemaRegistryWireFormatter.kt
```diff
@@ -0,0 +1,28 @@
...
+    @JvmStatic
+    fun addMagicBytes(bytes: ByteArray?): ByteArray {
+        if(bytes == null || bytes.isEmpty())
+            return ByteArray(0)
+
+        return kafkaSchemaRegistryMagicBytes() + bytes
+    }
+
+    private fun kafkaSchemaRegistryMagicBytes(): ByteArray {
+        val bytes = (0..3).map { 0x00.toByte() } + 0x01.toByte()
```
uglyog, line 25: This is always generating the magic bytes assuming the schema registry ID is 1. Would this be a problem if the ID is not 1?

uglyog (approved):

core/support/src/main/kotlin/au/com/dius/pact/core/support/json/KafkaSchemaRegistryWireFormatter.kt
muirandy: In short, no.

The Schema Registry aware Serializer/Deserializers will need to be configure with “mock:anything” to tell them not to talk to Schema Registry at all - for the tests. Having Schema Registry up and running would mean we’d also need Kafka (and likely Zookeeper)… and the abstraction that PACT has over the messaging layer means that we’re not actually writing to Kafka in the first place.

Configuring the [De]serializer with the mock will prevent it from trying to communicate to Schema Registry, but it will still read/write magic bytes.

We’re just adding on 5 bytes for the consumer, so that the Message(s) received contain 5 additional bytes. On the Producer, we’re expecting that the bytes produced will have 5 extra bytes at the start - so we’re starting parsing the JSON on the 6th byte.

uglyog: Sorry, the question was not about accessing the schema registry, but about the 5 bytes, which from the docs should be a 0 byte + 4 bytes for the registry ID. This PR hard codes the registry ID to 1. I don't have much knowledge in this area, just wondering if someone has interactions with two different registries (probably uncommon), would this be an issue?

muirandy: No worries! No - it won't be an issue. The 4 bytes represent an ID for a schema, but as we're not using the registry, they could be any value. In the admittedly unlikely event that there were 2 registries - only 1 could be used on a topic.

PR #1549, opened 2022-05-10
Upgrade build.gradle to mitigate Vulnerability CVE-2022-22965

rejeeshg:
Updating build.gradle to mitigate CVE-2022-22965 Vulnerability Issue


YOU54F:
Thanks for all these updates @rejeeshg, it hasn't gone unnoticed and is appreciated!

Not sure if you want to roll them all into branch, or are you preferring to do them incrementally in case of roll back?

If you haven't already, you can join our slack channel https://slack.pact.io

PR #1557, opened 2022-05-19
feat: Add support for providerVersionBranches

TGNThump:
Adds support for providerVersionBranches, aka providerBranches. Closes https://github.com/pact-foundation/pact-jvm/issues/1555


YOU54F:
It looks like CI is failing post this change going in, I've just merged in a doc change and noted master is failing, not had a chance to look into why ( I need a coffee and on a yak shaving trail but will add it to the list)

rholshausen:
Looks like I can't git.

PR #1565, opened 2022-05-28
fix: providerVersionBranch for pending pacts

praveen-em:
This fix is to address https://github.com/pact-foundation/pact-jvm/issues/1562


bethesque:
This looks like a breaking change. What is the release and communication process for this?

rholshausen:
It was only just added to the previous version (released 5 days ago), so not many people will be using it.

Issue #1569, opened 2022-06-09
Text content type override ignored and defaults to json anyway


fuudge:
Hey, I'm using below to override content type of application/graphql to text:
`System.setProperty("pact.content_type.override.application.graphql", "text")`

I can see the property being recognised, but I keep getting: 
`au.com.dius.pact.core.support.json.JsonException: Invalid JSON (1:3), found unexpected character 'd'`

'd' is the beginning of the graphql query which is not a valid JSON and why I want to treat it like a text. I've narrowed down the problem to the below line in ContentType.kt where it somehow matches the application/graphql to the json via superType.subtype which it gets from MediaTypeRegistry. 

[https://github.com/pact-foundation/pact-jvm/blob/d3580fccd1254e87263656133bf96bbcd15b0b0e/core/model/src/main/kotlin/au/com/dius/pact/core/model/ContentType.kt#L28](url)

is there anything I can do to stop this from being treated as json? i'm using the 4.3.8 version of pact-jvm

rholshausen:
I've just checked the [GraphQL spec](https://graphql.org/learn/serving-over-http/), the recommendation when using it over HTTP is to use JSON encoded bodies. Is this not the case for you?

What format is being used? If it is not JSON, then Pact will not be able to match it.

rholshausen:
Can you provide the full stack trace?

toomzzz:
I think it might be different for different language implementations. Here's the documentation for the GraphQL Java which shows a query syntax I'm using: https://www.graphql-java.com/documentation/batching
I think for the json queries the application/json could be used instead?

Here's the stack trace:
`au.com.dius.pact.core.support.json.JsonException: Invalid JSON (1:3), found unexpected character 'd'

	at au.com.dius.pact.core.support.json.JsonLexer.unexpectedCharacter(JsonParser.kt:88)
	at au.com.dius.pact.core.support.json.JsonLexer.nextToken(JsonParser.kt:79)
	at au.com.dius.pact.core.support.json.JsonParser.nextTokenOrThrow(JsonParser.kt:279)
	at au.com.dius.pact.core.support.json.JsonParser.parseObject(JsonParser.kt:184)
	at au.com.dius.pact.core.support.json.JsonParser.parse(JsonParser.kt:160)
	at au.com.dius.pact.core.support.json.JsonParser.parseString(JsonParser.kt:131)
[...]

Issue #1371, opened 2021-05-24
Request body encoded for graphql content type


fuudge:
Recently upgraded from Pact 3.6.2 to 4.1.21, everything looks ok except one test where graphql endpoint is used and the generated pact file uses encoded value for request body instead of actual graphql query. I'm guessing this isn't supported and is down to application/graphql being used instead of json, but is there any way to treat it like a text similar to json?

I'm using DSL to generate the RequestResponsePact and running with JUnit

fuudge:
I found the system property `pact.content_type.override.<TYPE>.<SUBTYPE>=text|binary|json` I can use to override the non binary "classification" of the graphql content type.

fuudge:
I found the system property `pact.content_type.override.<TYPE>.<SUBTYPE>=text|binary|json` I can use to override the non binary "classification" of the graphql content type.

uglyog:
Could you provide the content type that is used by GraphQL?

fuudge:
Sorry if that wasn't very clear but it's `application/graphql`

uglyog:
I'll add that to the list of json content types

PR #1570, opened 2022-06-09
fix: Don't parse graphql query as a JSON

toomzzz:
Related to issues #1569 and #1371 

In summary: original problem was request body was encoded when content type was application/graphql.
I found out you can always override the content type with property: pact.content_type.override.x.x 

The original ticket  #1371 was closed but the attempt to fix it made it so that the pact tries to parse graphql query as JSON, which it is not. i've raised that issue in #1569


rholshausen:
I'm closing this PR, as the default should be JSON (which is what most people will be using). This can be overwritten as needed.

PR #1605, opened 2022-09-16
Fix issue with consumer version selector method being protected

stefano-lucka:
Hi,

We've encountered an issue where the consumer version selector method was not picked up by the PACT library and only the `main` branch was checked against without noticing. After investigating I found out that a developer wrote this code: 

```
abstract class AbstractPactTest : AbstractBaseIntegrationTest() {
   
    /* Subclasses may override */
    @PactBrokerConsumerVersionSelectors
    protected fun consumerVersionSelectors(): SelectorBuilder = SelectorBuilder().branch("feature/feat-3996")
```
It doesn't work because the documentation clearly states that the method must be public. However an Exception is supposed to be thrown in this case which is not. People are now wondering why the feat-3996 PACT is not found. I figured that the consumer version selector was not used only by enabling debug logging of the "apache wire" package. 

It would be extremely helpful to always log out the actual used consumer version selector in a human readable way as INFO maybe, so that such issues would come to surface right away.

The reason the IllegalStateException doesn't throw is this: 

`java.lang.Class.getMethods()` 

> Returns an array containing Method objects reflecting all the public methods of the class or interface represented by this Class object, including those declared by the class or interface and those inherited from superclasses and superinterfaces.

Since method is `protected`,  `testClass?.methods` does not return it and validation logic is skipped, because no method is found. 
One solution is to use `getDeclaredMethods()` instead, however this one doesn't check the class hierarchy, so one needs to iterate the hierarchy oneself, which I did now.

I wrote a couple of unit tests and as you can see wasn't really eager to set this up in a super clean way (one should probably use a ParameterizedTest for that..). 

Since this is my first real contribution I hope the CICD runs through. Off for the weekend for now :-)

### DIFF
--- a/provider/junit5spring/src/test/java/au/com/dius/pact/provider/spring/junit5/PactBrokerLoaderTest.java
+++ b/provider/junit5spring/src/test/java/au/com/dius/pact/provider/spring/junit5/PactBrokerLoaderTest.java
@@ -1,0 +1,76 @@
+package au.com.dius.pact.provider.spring.junit5;
+
+import org.junit.jupiter.api.Test;
+
+import au.com.dius.pact.provider.junitsupport.loader.PactBrokerConsumerVersionSelectors;
+import au.com.dius.pact.provider.junitsupport.loader.PactBrokerLoader;
+import au.com.dius.pact.provider.junitsupport.loader.SelectorBuilder;
+
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+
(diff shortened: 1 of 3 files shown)


rholshausen:
Thanks for the PR! There were a few issues, which I fixed, but FYI:

* Static code analysis did not like all the unused inner test classes, just required some suppression
* The check for the method on the companion object was inside the loop over the declared methods. I moved it out.
* By returning the method from the companion object, you can't invoke it as it needs to be invoked with the companion object instance. I refactored the method to return both the method and the associated class so we can check for the companion object.

PR #1607, opened 2022-09-17
Use Thread Context ClassLoader to find test resources

holly-cummins:
Hi! I'm working on improving Pact support in Quarkus (see https://github.com/quarkusio/quarkus/issues/27729). At the moment, Pact provider tests works well when executed via a normal `mvn verify`, but do not work in Quarkus's dev mode. The reason is that in dev mode the classpath is less flat, and so it's no longer safe to assume the `PactLoader` is in the same classloader as the test classes. 

Pact can be made more tolerant to non-flat-classloader hierarchies by switching to use the Thread context classloader for finding resources. This should also improve behaviour in other frameworks.

### DIFF
--- a/provider/src/main/kotlin/au/com/dius/pact/provider/junitsupport/loader/PactFolderLoader.kt
+++ b/provider/src/main/kotlin/au/com/dius/pact/provider/junitsupport/loader/PactFolderLoader.kt
@@ -64,7 +64,7 @@ class PactFolderLoader : PactLoader {
   override fun getPactSource() = this.pactSource
 
   private fun resolvePath(): File {
-    val resourcePath = PactFolderLoader::class.java.classLoader.getResource(path.path)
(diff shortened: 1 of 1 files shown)


holly-cummins:
As an alternative implementation, we could look use the Pactloader's classloader, and then fallback to the thread context classloader if there's nothing found using Pact's loader; or the other way round.

rholshausen:
All the things I can find about the Thread context classloader online say not to use it, but that is for class loading (makes sense, because if the thread changes, then what classes are accessible can change). I think it is safe for test resources, because the thread should not change during the execution of a test.

rholshausen:
Now that I think about it, the problem may be that it is using `PactFolderLoader::class.java.classLoader`, when it should be using `getClass().getClassLoader()`.

holly-cummins:
Thanks for the fast merge, @rholshausen!

Issue #1632, opened 2022-11-02
Custom authentication header for PactBroker access


artemptushkin:
Having Pact Broker under an authentication layer with a custom header, like `my-authorization`  instead of `Authorization` makes it impossible to propagate the value to the client

Is there a way to propagate a custom header for Pact Broker access?

Relevant for: Maven, Gradle plugins and provider clients

If no, can this feature be implemented?

artemptushkin:
@rholshausen is it worth contributing a feature to customize the `Authorization` header key with a custom value from the properties?

artemptushkin:
closed with #1633 

the default auth header can be customized now from the `Authorization` through any library using a dedicated property (e.g. `pactBrokerAuthenticationHeader`), see corresponded documentation for more

PR #1633, opened 2022-11-06
GH-1632 invent default auth header with additional tests

artemptushkin:
Having the Pact broker located behind a custom auth layer with a custom header but `Authorization` it is not possible to access the broker, the PR is a proposal to customize it


artemptushkin:
@rholshausen thank you for the merge! can you release it, please? or give an approximate time when it can be published

rholshausen:
4.3.17 is released

PR #1645, opened 2022-12-06
bugfix: fix custom header propagation in list case

artemptushkin:
I forgot to propagate it here, it's not a nice place to be covered by tests


artemptushkin:
@uglyog can you please release when merged, it's tested on my private project and we're waiting for this. Thank you

PR #1663, opened 2023-02-05
Breaking Change - Support SpringBoot 3 & Java 17

### LINKED ISSUE #1660: java.lang.NoSuchMethodError due to changed return type in org.springframework.test.web.reactive.server.EntityExchangeResult.getStatus()
We've switched to Spring Boot 3 in conjunction with Spring Framework 6.

While verifying with…
```
    @TestTarget
    val target = WebFluxTarget(router.manufacturer())

    @TestTemplate
    @ExtendWith(PactVerificationSpringProvider::class)
    fun pactVerificationTestTemplate(context: PactVerificationContext) {
        context.verifyInteraction()
    }
```
…we are now experiencing the following error…
`java.lang.NoSuchMethodError: 'org.springframework.http.HttpStatus org.springframework.test.web.reactive.server.EntityExchangeResult.getStatus()'`

If you compare [ExchangeResult in Spring Framework 5](https://github.com/spring-projects/spring-framework/blob/5.3.x/spring-test/src/main/java/org/springframework/test/web/reactive/server/ExchangeResult.java#L178) it still returns `HttpStatus` while [in Spring Framework 6](https://github.com/spring-projects/spring-framework/blob/main/spring-test/src/main/java/org/springframework/test/web/reactive/server/ExchangeResult.java#L175) it returns `HttpStatusCode`.

Can you adapt to this new reality?

ludorival:
Hello,

Here is a PR to support Spring Boot 3 on Pact JVM. Since SpringBoot 3 is compiling with JDK 17, I had to pass to Java 17 in the project as well.

I have been able to test from another project  and it is working with Spring Boot 3 and this local version `provider:junit5spring:5.0.0`.

This PR should fix #1660 

Feel free to review and provide your suggestions.

Thanks,
Ludo

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -10,7 +10,7 @@
 
 Most of Pact-JVM is written in Kotlin and is built with Gradle. Tests are written using [Spock](https://spockframework.org/).
 
-Before you build, install java 11, `Gradle` and `Maven` (Maven is required to build the Maven plugin).
+Before you build, install java 17, `Gradle` and `Maven` (Maven is required to build the Maven plugin).
 
 #### To build the libraries:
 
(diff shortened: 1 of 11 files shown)


rholshausen:
Thanks for the PR, the change looks simple enough. I might rather add a new spring6 module that requires JDK 17 instead of making all the modules require that version.

ludorival:
Good catch @rholshausen, I can set jdk 7 for spring modules only ?

rholshausen:
The changes look quite simple (just some imports that are different). I am thinking of adding it to a new spring6 module, leaving the existing ones as is. Also, I don't think it has to be compiled with JDK 17, just run with it. 

I will set this all up and move your changes there, I just need a little bit of time to setup a Springboot project to test it works ok.

rholshausen:
Ok, I have taken your changes and put them in a new spring6 module on the v4.5.x branch. I've tested it with a demo Springboot3 project, and it works ok. Commit bffd26aba8a9cc18ec543dd9c2cd3b17182e01c6

ludorival:
Awesome @rholshausen ! Can I close this PR then ?

rholshausen:
Yes, closing this as 4.5.0-beta1 has been released with the new spring6 lib

PR #1670, opened 2023-02-24
fix merging message pacts - old messages taking precedence

IlyaNerd:
So when we change our pact to contain new field, for example, we run the consumer test and it doesnt update the pact file because it thinks that the message is the same.
The bug being is that we adding 2 collections - `old + new` - and then `distinctBy` - in this case old duplicates will take precedence


corbym (approved):
I approve.

IlyaNerd:
hey @rholshausen, could we include this into 4.4.x pls?

rholshausen:
It will be in the next 4.4.x release

PR #1685, opened 2023-04-19
Use shorthand annotation instead of @ExtendWith

holly-cummins:
What do folks think of this? 

Instead of 

```java
@ExtendWith(PactConsumerTestExt.class)
class ExampleJavaConsumerPactTest {
```
... a shorter form:

```java
@PactConsumerTest
class ExampleJavaConsumerPactTest {
```

It has the disadvantage that 
- Now there's two ways to do things
- It's less obvious what's happening under the covers, to people who know JUnit 5
- If JUnit 6 (say) introduced a totally different API, the `@PactConsumerTest` annotation would have to change, even though on the surface it looks like it should be independent of JUnit versions 
- There's a bit of overlap with the meaning of `@PactTestFor`. I wondered about just adding the `@ExtendWith` to `@PactTestFor` but I don't think it works since it can also be used on methods

It has the advantage that
- It's tidier and easier to read, with less typing
- It works more easily with tab completion

I've just done it for the consumer side, but if the community thinks it's a good idea, we'd want it for both consumer and provider.

### DIFF
--- a/consumer/junit5/README.md
+++ b/consumer/junit5/README.md
@@ -15,9 +15,15 @@ The library is available on maven central using:
 
 ### 1. Add the Pact consumer test extension to the test class.
 
-To write Pact consumer tests with JUnit 5, you need to add `@ExtendWith(PactConsumerTestExt)` to your test class. This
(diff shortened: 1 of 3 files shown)


rholshausen:
Nice!

PR #1691, opened 2023-05-12
Allow to set the port via an expression.

holomekc:
At the moment it is not possible to define the used port via an expression. I am aware that there is the 0 = random port, but we need this because the port needs to be set in a process, which is running way before the random port of pact is calculated. So for us we have a kind of "dirty" workaround, which determines a random port, sets this to the process I mentioned and then applies this port to pact. It would be great to set the port via a system property.

What do you think?

Br,
Chris


rholshausen:
Thanks for the PR!

PR #1694, opened 2023-05-22
use valid module name in manifest

Danny02:
commit https://github.com/pact-foundation/pact-jvm/commit/0338f438c3d4fc82c49e6c88c4833cb44238af30 (see settings.gradle) broke the previouse work around


rholshausen:
Ah! That explains a few things. I might also revert this
image

PR #1703, opened 2023-07-24
upgrade to kotlin-logging 5

oshai:


oshai:
Hi, I created a PR to upgrade kotlin-logging to v5, hope you will be open to accept it. I am the lib maintainer and since this is a breaking change and pact is pretty popular I thought to assist with the upgrade.

Locally the project is compiling now, if you can run the workflow it will help to make sure also in CI.
Notable changes:
- replace `import mu.` with `import io.github.oshai.kotlinlogging.`
- add dependency `api 'io.github.oshai:kotlin-logging-jvm:5.0.0'`. For now I kept the old dependency as a side by side, can be removed here or later to avoid breaking consumers relying on that.

What's left to do: Some methods and classes are deprecated (but still works as before). So I can change that as well in this cl, or later.

rholshausen:
Awesome, thanks for the PR!

PR #1706, opened 2023-07-28
fix: generate providerStates as in V3 when PactVersion is UNSPECIFIED

victorenaud:
fix https://github.com/pact-foundation/pact-jvm/issues/1705

I chose the path of least modifications.
A more definitive solution might be to change the default pactVersion to V3 in `PactSpectVersion` instead of `UNSPECIFIED`, but that would change the philosophy of how it is done.


rholshausen:
I'm thinking that, rather than changing the order of the enum values, I'm going to remove `UNSPECIFIED` and make `PactSpectVersion` and optional value. That way it will be unspecified when there is no value, and then the default choice can be made as necessary.

rholshausen:
Fixed with bd2f380

PR #1713, opened 2023-09-04
feat: Retry all http methods

monochromata:
https://github.com/pact-foundation/pact-jvm/issues/1710


rholshausen:
Thanks for the PR

PR #1744, opened 2023-12-08
Implemented pact with plain java and Junit5

PrudhviRajMukiri23:
In this PR, I have implemented both consumer and provider with plain java and Juint5. Also added steps to setup pact server in local using docker composer.


rholshausen:
The tests as they are cause the build in CI to fail.

PR #1815, opened 2024-07-14
POC: Test Templates for Async Message Tests

timvahlbrock:


timvahlbrock:
Aside from the code style and a major change in Pact method detection the main issue is that JUnit will currently try to inject the message as a parameter directly, instead of the message context, which will cause the same message being used two times (implemented an assertion to test this).

timvahlbrock:
Maybe JUnit is not able to handle a Parameter provider and a TestTemplateContext Provider on the same extension?

rholshausen:
I'm getting a lot of test failures with this change. The issue (as you have observed) is that with a test template, there is no test instance. There is lots of places where it tries to resolve something on the test instance.

PR #1822, opened 2024-08-28
Make plugindriver version consistent

huehnerlady:
In our project we have the following config:

```
configurations {
    all {
      resolutionStrategy {
        failOnVersionConflict()
        force(dependencyVersions)
      }
    }
  }

```

With that we see that with the update to pact 4.6.14 there is a new version conflict of the plugin driver

Looking in the project I can see, that there is `io.pact.plugin.driver:core:0.5.1` and `io.pact.plugin.driver:core:0.4.2` defined, depending if you use `api` or `implementation`.

I aligned the version now. In our projects we have either reminder comments or we have variables.
I here thought the least intrusive way is a comment to hopefully remind people to update the version in 2 places in future.


rholshausen:
Oops! Thank for the PR!

PR #1858, opened 2025-03-16
fix: Ensure provider state is injected when verifying sync message pacts

WellingR:
When working with synchronous message pact tests we were running into an issue where we noticed that the provider state was not being injected in the request messages.

I did some investigation and it seems that this was just not done for sync message pacts.
This PR adds a change to make sure that the generated request is included in the interaction.

### DIFF
--- a/provider/src/main/kotlin/au/com/dius/pact/provider/ProviderVerifier.kt
+++ b/provider/src/main/kotlin/au/com/dius/pact/provider/ProviderVerifier.kt
@@ -1451,6 +1451,7 @@ open class ProviderVerifier @JvmOverloads constructor (
           m.invoke(instance)
         }
       } catch (e: Throwable) {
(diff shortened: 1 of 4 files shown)


provider/src/main/kotlin/au/com/dius/pact/provider/ProviderVerifier.kt
```diff
@@ -1451,6 +1451,7 @@ open class ProviderVerifier @JvmOverloads constructor (
           m.invoke(instance)
         }
       } catch (e: Throwable) {
+        logger.warn(e) { "Failed to invoke provider method '${m.name}'" }
```
WellingR, line 1454: This may seem unnessecary, but the cause/stacktrace of the original throwable is not being logged otherwise

PR #1887, opened 2025-12-13
Upgrade tika-core 2.9.4 > 3.2.3 for issue 1885

csbiggar:
The current tika-core version has vulnerabilities as per https://github.com/pact-foundation/pact-jvm/issues/1885

Upgrading tika from 2.x to 3.x  includes some breaking changes as per their [release notes](https://github.com/apache/tika/blob/main/CHANGES.txt) 

The following breaking changes showed up in this code via test failures in ContentTypeSpec and have been fixed:

> Tika will look for "custom-mimetypes.xml" directly on the classpath, NOT under "/org/apache/tika/mime/". (TIKA-4147).

* This is a simple file move and seems to work fine (according to the ContentTypeSpec unit tests)

> Return media type "text/javascript" instead of "application/javascript" to follow RFC-9239. (TIKA-4119).
*  I chose to change the test expectation to honour this rather than try to force backward compatibility - is that the right thing to do? I think that means the pact version this is included in will be a breaking change? An alternative could be to add this to the custom-mimetypes.xml


csbiggar:
Thanks! Do I need to do any further admin to get it included in the next release (is there a release planned?)

